PR #5, opened 2018-05-27
Fixes so snet command will install and run in Windows

ferrouswheel:
I haven't tested each command. Just checked that the `snet` command could be found and that it would prompt for an identity.

To avoid x-platform issues in future, we should try to use `entry_points` option in the setup call.

It also bumps the patch number to `0.1.3`, but I can remove that if there's a specific version release process we're following.


lupuleasa-core:
Can the SNET cli be used on windows?

I installed it in my conda environment, using pip

but I don't know how to install libudev and libusb

are they supported on windows?

ferrouswheel:
The idea is to support most platforms for the client and SDK, and at the time I raised this it was decided we should support windows. Lots has changed since then tho...

Can you open a new issue with the specific issue around libudev and libusb?

lupuleasa-core:
@ferrouswheel 

I just don't know how I should install the two on windows.

Is the CLI installable only by running pip3 install snet-cli?

Any other operations that must be done on windows?

I installed it in my conda env, and then I tried to call "snet" in the cmd and it didn't work

I figured it was some path issues, then I searched in and found snet.exe in my script folder and when I ran it it didn't work, gave an error saying can't find module snet-cli

ferrouswheel:
Discussion will continue in #227

PR #37, opened 2018-09-13
get-model and updating current_agent_address after service publish

arturgontijo:
[fix] `snet service client get-model dest_dir/` if `dest_dir/` already exists.
[add] Updating `current_agent_address` after `snet service publish` so the user can run `snet client call` right after the publication, avoiding manually set `current_agent_address`.

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -449,9 +449,9 @@ class ClientCommand(BlockchainCommand):
             model_dir = self._getstring("dest_dir") or Path("~").expanduser().joinpath(".snet").joinpath("models").joinpath(model_hash)
             if not os.path.exists(model_dir):
                 os.makedirs(model_dir)
-                model_tar = ipfs_client.cat(model_hash)
-                with tarfile.open(fileobj=io.BytesIO(model_tar)) as f:
-                    f.extractall(model_dir)
+            model_tar = ipfs_client.cat(model_hash)
+            with tarfile.open(fileobj=io.BytesIO(model_tar)) as f:
+                f.extractall(model_dir)
             self._pprint({"destination": str(model_dir)})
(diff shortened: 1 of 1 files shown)


snet_cli/commands.py
```diff
@@ -449,9 +449,9 @@ def get_model(self):
             model_dir = self._getstring("dest_dir") or Path("~").expanduser().joinpath(".snet").joinpath("models").joinpath(model_hash)
             if not os.path.exists(model_dir):
                 os.makedirs(model_dir)
-                model_tar = ipfs_client.cat(model_hash)
-                with tarfile.open(fileobj=io.BytesIO(model_tar)) as f:
-                    f.extractall(model_dir)
+            model_tar = ipfs_client.cat(model_hash)
+            with tarfile.open(fileobj=io.BytesIO(model_tar)) as f:
+                f.extractall(model_dir)
```
vsbogd: May be we should remove all previous data to clear space for new model. What do you think?

arturgontijo: It's a good idea, if you see the line 449 if "dest_dir" isn't provided the IPFS model hash is concatenated with ".snet/models" to separate different .proto from services.
Maybe a better idea is to do this with "dest_dir", and then remove a previous IPFS model hash before get the new one.
But I'll have to analyse the impact of it when publishing/updating the service.

vsbogd: Thanks - I didn't pay attention at line 449. I think is ok for now then.

arturgontijo: For now, I think we must not remove files/folder with this command.

vsbogd: But probably not in case when ```dest_dir``` is omitted.

Issue #36, opened 2018-09-12
Change service protobuf folder name model/ to service_spec/


arturgontijo:
As mentioned on Slack, the folder name `model/` could cause confusion with trained NN models of a Service.

arturgontijo:
Fixed #45

PR #39, opened 2018-09-13
Change reference from "service model" to "service spec"

arturgontijo:
Fix issue: #36


arturgontijo:
Changing `service-spec /` to `service_spec /` to avoid python SyntaxError when importing the folder within the service.

Issue #40, opened 2018-09-13
Create service delete


arturgontijo:
- Create a high level interaction with Registry to delete a service.

PR #43, opened 2018-09-13
Create service delete

arturgontijo:
Related to issue #40.

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -449,9 +449,9 @@ class ClientCommand(BlockchainCommand):
             model_dir = self._getstring("dest_dir") or Path("~").expanduser().joinpath(".snet").joinpath("models").joinpath(model_hash)
             if not os.path.exists(model_dir):
                 os.makedirs(model_dir)
-                model_tar = ipfs_client.cat(model_hash)
-                with tarfile.open(fileobj=io.BytesIO(model_tar)) as f:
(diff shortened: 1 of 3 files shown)


snet_cli/commands.py
```diff
@@ -1224,3 +1228,65 @@ def update(self):
...
+
+        agent_address = service_json.get('networks', {}).get(network_id, {}).get('agentAddress', None)
+        if agent_address is None:
+            self._error("Service hasn't been deployed to network with id {}".format(network_id))
+
+        self._printout("Getting information about the service...")
+        registry_contract_def = get_contract_def("Registry")
+        registry_address = self._getstring("registry_at")
+
+        (found, _, current_path, current_agent_address, current_tags) = ContractCommand(
```
vsbogd: I saw the similar code in previous PR: https://github.com/singnet/snet-cli/pull/41, https://github.com/singnet/snet-cli/blob/5e64057c1a6984dac4191113233e1b2750e6e7a4/snet_cli/commands.py#L1084
I would remove code duplication and create some python method to call this ```getServiceRegistrationByName()```. Actually may be it makes sense to wrap all contract calls by Python methods to not copy/paste these long code snippets. What do you think?

arturgontijo: Agree!
I did something similar for the `organization` command.
I'll do the same for the `service`.

Issue #42, opened 2018-09-13
Organization command


arturgontijo:
- Create `snet organization [arguments]`
- User will be able to interact with the Organization's functions in a high level.
- List of available low level functions (high level `[arguments]`):
    - createOrganization (`create`)
    - changeOrganizationOwner (`change-owner`)
    - addOrganizationMembers (`add-members`)
    - removeOrganizationMembers (`rem-members`)
    - deleteOrganization (`delete`)
    - listOrganizations (`list`)
    - getOrganizationByName (`info`)
    - listServicesForOrganization (`list-services`)
    - listTypeRepositoriesForOrganization (`list-repo`)

PR #44, opened 2018-09-14
Organization command

arturgontijo:
Related: #42

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -1284,3 +1284,29 @@ class OrganizationCommand(BlockchainCommand):
             if not self.args.members:
                 self._printerr("\nMissing Organization's members: --members ORG_MEMBERS[]\n")
             self._error("Missing params...")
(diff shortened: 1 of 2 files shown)


snet_cli/commands.py
```diff
@@ -1225,3 +1225,324 @@ def update(self):
...
+                try:
+                    cmd.transact()
+                except Exception as e:
+                    self._printerr("\nTransaction error!\nHINT: Check if {} already exists.\n".format(self.args.name))
+                    self._error(e)
+            except Exception as e:
+                self._printerr("\nTransaction error!\nHINT: Check if address is in hexadecimal with 32 chars.\n".format(self.args.name))
+                self._error(e)
+        else:
+            self._error("Missing params...")
```
vsbogd: I would mark required parameters as required when calling ```add_argument``` method. It is not clear which parameters are missing otherwise.

PR #47, opened 2018-09-17
Create service delete

arturgontijo:
Related: #40


vsbogd:
Thanks!

BTW, I know this code doesn't seems easy to write unit tests but I think we need to try ). Do you know any good Python testing framework for Ethereum? Just googled and found https://github.com/ethereum/eth-tester.

arturgontijo:
No, I don't. But I agree with your.
I'll check this `eth-tester` and see how it works!

vsbogd:
Great, thank you!

PR #50, opened 2018-09-18
Organization command

arturgontijo:
Related: #42

### DIFF
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -78,7 +78,7 @@ def add_root_options(parser, config):
     add_service_options(service_p, config)
 
     organization_p = subparsers.add_parser("organization", help="Interact with SingularityNET Organizations")
-    add_organization_options(organization_p, config)
+    add_organization_options(organization_p)
 
 
 def add_identity_options(parser, config):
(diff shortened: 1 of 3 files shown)


snet_cli/arguments.py
```diff
@@ -325,6 +333,187 @@ def add_service_options(parser, config):
...
+    _add_organization_arguments(p)
+
+    org_create_p = subparsers.add_parser("create", help="Create an Organization", default_choice="default")
+    org_create_p.set_defaults(fn="create")
+    networks_organization_subparsers = org_create_p.add_subparsers(title="networks", metavar="[NETWORK]")
+    org_create_p.add_argument("name", help="Name of the Organization", metavar="ORG_NAME")
+    org_create_p.add_argument("members", help="List of members to be added to the organization",
+                              metavar="ORG_MEMBERS[]")
+
+    for network_name in network_names:
```
vsbogd: As far as I understand for each organization command the list of networks is added as additional options for 
the command. On the other hand root parser already has --network option to specify the network. Can it be used here?

vsbogd:
It is worth to update https://github.com/singnet/snet-cli/blob/master/README.md to add new commands.

snet_cli/arguments.py
arturgontijo: Yes, I think we can get `network` from `config` (session).
I'll change it and make some tests to see if it's OK.

vsbogd:
Looks good for me, have finished the tests you mentioned at https://github.com/singnet/snet-cli/pull/50#discussion_r218894835 ?

arturgontijo:
Yes but only at `argparse` level.
I didn't use [eth-teste](https://github.com/ethereum/eth-tester) yet.

Issue #53, opened 2018-09-24
Error message is not printed when  service spec retrieving is failed


stellarspot:
Steps to reproduce:
- Go to the example-service project
- Run in command line:
`snet service init`
- For the choice 'Choose the path to your service's spec directory: (default: "service_spec/")' type . (current directory, this allows to resolve the provided path)
- Use snet to publish the local config
-Run example-service with snet-daemon
- Execute  snet client call classify 

The error 'ERROR: failed to retrieve service spec' is not printed.

arturgontijo:
I was able to reproduce this as well.
Maybe we could specify the folder in ERROR: message, what do you think?

Like: 'ERROR: failed to retrieve service spec files from . (or whatever user has specified at service init)'

arturgontijo:
The folder specification with ERROR won't make sense for third party service...
If we pass `out_f=self.err_f, err_f=self.err_f,` at the get_spec() call, the message will be printed.
At [command.py:L510-L516](https://github.com/singnet/snet-cli/blob/master/snet_cli/commands.py#L510-L516)

PR #55, opened 2018-09-24
Error message when fail to retrieve service spec

arturgontijo:
#53

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -510,8 +510,8 @@ class ClientCommand(BlockchainCommand):
         spec_hash = ClientCommand(
             config=self.config,
             args=self.args,
-            out_f=None,
-            err_f=None,
+            out_f=self.err_f,
+            err_f=self.err_f,
             w3=self.w3,
             ident=self.ident).get_spec()
 


snet_cli/commands.py
```diff
@@ -510,8 +510,8 @@ def call(self):
         spec_hash = ClientCommand(
             config=self.config,
             args=self.args,
-            out_f=None,
-            err_f=None,
+            out_f=self.err_f,
```
stellarspot: Why self.err_f is used for  out_f instead of self.out_f?

stellarspot: There are other places where out_f and err_f are set to None.
For example ContractCommand: https://github.com/singnet/snet-cli/blob/master/snet_cli/commands.py#L302
Should these places updated as well?

arturgontijo: My guess is that the previous dev team use `self.out_f` (stdout) only for output transaction data.
For "basic" logging they use the `self.err_f` (stderr).
Like they did [here](https://github.com/singnet/snet-cli/blob/841b6d3bc3a2fa9c5b0856ae2ecb4a732b4ee949/snet_cli/commands.py#L589-L590)

arturgontijo: > There are other places where out_f and err_f are set to None.
> For example ContractCommand: https://github.com/singnet/snet-cli/blob/master/snet_cli/commands.py#L302
> Should these places updated as well?

I think so...cause I prefer more logs then less.
The downside is that will be a bunch of logs from "intermediary" transactions.
Maybe we can use this in a `--verbose` mode.

stellarspot: It sounds good. It would be better to move it into a separate issue.

Issue #54, opened 2018-09-24
Transaction with "blockHash: null"


arturgontijo:
Sometimes the transaction returns this receipt:

```
Proceed? (y/n): y
Submitting transaction...

    event_summaries:
    -   args:
            agent: '0xFD55Dc4276E31C7F2022997A2e8F296043C26D7b'
        event: AgentCreated
    receipt_summary:
        blockHash: null
        blockNumber: null
        cumulativeGasUsed: 2878039
        gasUsed: 1788285
        transactionHash: '0x22b8c631a1e9be622d6da1187538e940b983595f15ce966331f59e57ef6f9826'
```

Checking:
[identity.py:58-61](https://github.com/singnet/snet-cli/blob/master/snet_cli/identity.py#L58-L61)
[identity.py:86-89](https://github.com/singnet/snet-cli/blob/master/snet_cli/identity.py#L86-L89)
[identity.py:128-131](https://github.com/singnet/snet-cli/blob/master/snet_cli/identity.py#L128-L131)
[identity.py:179-182](https://github.com/singnet/snet-cli/blob/master/snet_cli/identity.py#L179-L182)
[identity.py:291-294](https://github.com/singnet/snet-cli/blob/master/snet_cli/identity.py#L291-L294)
I saw that it just wait for something in receipt but it doesn't check the content of the fields insite it.

Suggestion:
```
# Wait for transaction to be mined
receipt = dict()
while not receipt:
    time.sleep(1)
    receipt = self.w3.eth.getTransactionReceipt(txn_hash)
    if receipt and "blockHash" in receipt and receipt["blockHash"] is None:
        receipt = dict()
```

PR #57, opened 2018-09-24
Better "Wait for transaction to be mined"

arturgontijo:
Avoiding get `blockHash: null` at transaction receipt.
Related: #54

### DIFF
--- a/snet_cli/identity.py
+++ b/snet_cli/identity.py
@@ -55,10 +55,12 @@ class KeyIdentityProvider(IdentityProvider):
         txn_hash = self.w3.eth.sendRawTransaction(raw_transaction)
 
         # Wait for transaction to be mined
-        receipt = None
-        while receipt is None:
+        receipt = dict()
+        while not receipt:
             time.sleep(1)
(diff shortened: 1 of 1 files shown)


vsbogd:
@astroseger, could you please review it?

astroseger:
There are 5 places with the same code. I would propose to make a function: send_and_wait_for_transaction(raw_transaction).
Also I have a question: what happens if transaction never mined (for example gas price was way too low, or something else)?

arturgontijo:
I agree to create a function.
About the never ending case, the previous code didn't implement a timeout but I think it's necessary too.
How long? 30s, 60s?
According to [BlockTime](https://etherscan.io/chart/blocktime), it'll be ok with 60s.

astroseger:
On main ethereum net you can wait much longer then 60s. In any case, a simple timeout is not enough, because transaction can be mined after timeout. So I would propose to leave it as it is for now (without timeout). And later, we probably will need to develop a protocol to deal with transactions which take too much time. We should make sure that transaction was not mined because of the low gas price, and not because of some other reasons, and than resend it with the same nonce but higher gas price...

Issue #52, opened 2018-09-21
Publish service with same contract address is currently allowed.


arturgontijo:
If you publish a service using a service.json and then publish it again with different name a new service will be published at the same Contract Address.

```
(snet_015) artur@singNET:~/Myservice$ cat service.json
{
    "name": "Service1",
    "service_spec": "service_spec/",
    "organization": "SNET_BH",
    "path": "",
    "price": 2,
    "endpoint": "http://54.203.198.53:7777",
    "tags": [
        "Tag_1,",
        "Tag_2,",
        "Tag_3"
    ],
    "metadata": {
        "description": "Testing..."
    },
    "networks": {
        "42": {
            "agentAddress": "0xc3d631F11A3cf5628d8bb52EC12EDB654e579F3c"
        }
    }
}
(snet_015) artur@singNET:~/Myservice$ snet service publish
Creating transaction to update agent contract's metadataURI from ipfs:/ipfs/QmZWchjSCqbLnbg8rMajdTsGudRP9VC4dhvUoSMSqd3nZo to ipfs:/ipfs/QmaFu326ZyEsYoe3NzghUrX9f7Ycq2fbX4cW6nFWTkKdeV...

    transaction:
        chainId: 42
        data: '...'
        from: '0xFF2a327ed1Ca40CE93F116C5d6646b56991c0ddE'
        gas: 42081
        gasPrice: 1000000000
        nonce: 2650
        to: '0xc3d631F11A3cf5628d8bb52EC12EDB654e579F3c'
        value: 0

Proceed? (y/n): y
Submitting transaction...

    event_summaries: []
    receipt_summary:
        blockHash: '0x398dcba46ec7fefd24b522283e9552994e6ff2d94864cc0cb1719d5ab547e622'
        blockNumber: 8804199
        cumulativeGasUsed: 173237
        gasUsed: 42081
        transactionHash: '0xe3e10bb674bdda400f23a43bb9498f7253b813aed2ca240a64ab1b3a32a73b75'

Creating transaction to update agent contract's endpoint from http://54.203.198.53:7000 to http://54.203.198.53:7777...

    transaction:
        chainId: 42
        data: '...'
        from: '0xFF2a327ed1Ca40CE93F116C5d6646b56991c0ddE'
        gas: 34930
        gasPrice: 1000000000
        nonce: 2651
        to: '0xc3d631F11A3cf5628d8bb52EC12EDB654e579F3c'
        value: 0

Proceed? (y/n): y
[...]

Issue #49, opened 2018-09-17
Publishing service with invalid endpoint


arturgontijo:
- Endpoints:
  - test:?/><%$endpoint
  - http://test?&^*)(.com

vsbogd:
@arturgontijo , could you please explain it further, it is not clear how to reproduce it.

arturgontijo:
#48

Issue #48, opened 2018-09-17
Publishing service with invalid .proto file


arturgontijo:
- Service with plain text instead of protobuf (.proto) file.
  - Service with ~1Mb file:
    -rw-rw-r-- 1 artur artur 1013076 Set 17 17:04 NOT_PROTO_FILE_1MB.proto
  - But not with > 1.5Mb file:
    -rw-rw-r-- 1 artur artur 1475526 Set 17 16:59 NOT_PROTO_FILE_1_5MB.proto

vsbogd:
@arturgontijo , could you please explain it further, it is not clear how to reproduce it.

arturgontijo:
```
(snet_015) artur@singNET:~/invalid_proto$ ls -la service_spec/
total 1000
drwxrwxr-x 2 artur artur    4096 Set 18 07:34 .
drwxrwxr-x 3 artur artur    4096 Set 18 07:34 ..
-rw-rw-r-- 1 artur artur 1013076 Set 18 07:34 NOT_PROTO_FILE_1MB.proto

(snet_015) artur@singNET:~/invalid_proto$ snet service init
Please provide values to populate your service.json file

Choose a name for your service: (default: "invalid_proto")

Choose the path to your service's spec directory: (default: "service_spec/")

Choose an organization to register your service under: (default: "")
SNET_BH
Choose the path under which your Service registration will be created: (default: "")

Choose a price in AGI to call your service: (default: 0)
1
Endpoint to call the API for your service: (default: "")
akdfhasjfhaksdhfakjhfl
Input a list of tags for your service: (default: [])
1,2,3 
Input a description for your service: (default: "")
Description...
{
    "name": "invalid_proto",
    "service_spec": "service_spec/",
    "organization": "SNET_BH",
    "path": "",
    "price": 1,
    "endpoint": "akdfhasjfhaksdhfakjhfl",
    "tags": [
        "1,2,3"
    ],
    "metadata": {
        "description": "Description..."
    }
}

service.json file has been created!

(snet_015) artur@singNET:~/invalid_proto$ snet service publish
Creating transaction to create agent contract...

    transaction:
        chainId: 42
        data: '...'
        from: '0xFF2a327ed1Ca40CE93F116C5d6646b56991c0ddE'
        gas: 1788267
        gasPrice: 1000000000
        nonce: 2532
        to: '0x1fAa8ec70aFe4f5ce904dA935A6ddF5f3482eEDb'
        value: 0

[...]

PR #60, opened 2018-09-24
Checking .proto, price/endpoint in service json and Agent Contract

arturgontijo:
- Remove some newlines (code structure).
- Exception at all `transact()` in `publish()`.
- All found .proto are compiled to check if they are valid before publish/update (Related #48).
- `snet` checks if `price` field is a valid number.
- `snet` sends a GET request to endpoint before publish/update (Related #49).
- Check the current state of the Agent Contract [L990-L1009] (Related #52).

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -9,22 +9,20 @@ import tempfile
 from pathlib import Path
 from shutil import rmtree
 from textwrap import indent
-from urllib.parse import urljoin, urlparse
+from urllib.parse import urljoin
 
 import grpc
 import ipfsapi
-import pkg_resources
 import requests
 import yaml
 from google.protobuf import json_format
-from grpc_tools.protoc import main as protoc
 from rfc3986 import urlparse, uri_reference
 
 from snet_cli.contract import Contract
 from snet_cli.identity import get_kws_for_identity_type, get_identity_types
 from snet_cli.session import from_config, get_session_keys
 from snet_cli.utils import DefaultAttributeObject, get_web3, get_identity, serializable, walk_imports, \
-    read_temp_tar, type_converter, get_contract_def, get_agent_version
+    read_temp_tar, type_converter, get_contract_def, get_agent_version, compile_proto
 
 
 class Command(object):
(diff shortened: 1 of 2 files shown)


snet_cli/commands.py
```diff
@@ -813,6 +812,102 @@ def _getserviceregistrationbyname(self, service_json=None):
...
+                entry_path = Path.cwd().joinpath(spec_path).resolve()
+            else:
+                entry_path = spec_path.resolve()
+            if not os.path.isdir(entry_path):
+                self._printerr("Service spec path must resolve to a valid directory: {}".format(spec_path))
+                is_ok = False
+            import_paths = walk_imports(entry_path)
+
+            if import_paths:
+                codegen_tmp_dir = Path("~").expanduser().joinpath(".snet").joinpath("py-codegen").joinpath(
```
vsbogd: I would use [tempfile.mkdtemp](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp) instead

snet_cli/commands.py
```diff
@@ -813,6 +812,102 @@ def _getserviceregistrationbyname(self, service_json=None):
...
+        metadata_ipfs_uri = uri_reference(metadata_ipfs_path).copy_with(scheme='ipfs').unsplit()
+
+        # Checking price
+        if not type(service_json["price"]) == int:
+            self._printerr("Invalid price format: {} [{}]".format(service_json["price"], type(service_json["price"])))
+            is_ok = False
+
+        # Checking endpoint
+        try:
+            request = requests.get(str(service_json["endpoint"]), timeout=5)
```
vsbogd: Do not think we should actually check that endpoint is live as it prevents us from publishing service which is temporary down. Is it enough to check that endpoint has correct URI format?

snet_cli/commands.py
vsbogd: BTW it seems this code duplicates https://github.com/singnet/snet-cli/blob/fdbaea78e566c6d470996a3243fe288f9532a94d/snet_cli/commands.py#L518

snet_cli/commands.py
```diff
@@ -813,6 +812,102 @@ def _getserviceregistrationbyname(self, service_json=None):
...
+                # stdout back to console
+                sys.stdout = sys.__stdout__
+                if os.path.exists(codegen_tmp_dir):
+                    rmtree(codegen_tmp_dir)
+
+        # Create service_spec/ tar and upload it to IPFS
+        ipfs_endpoint = urlparse(self.config["ipfs"]["default_ipfs_endpoint"])
+        ipfs_scheme = ipfs_endpoint.scheme if ipfs_endpoint.scheme else "http"
+        ipfs_port = ipfs_endpoint.port if ipfs_endpoint.port else 5001
+        ipfs_client = ipfsapi.connect(urljoin(ipfs_scheme, ipfs_endpoint.hostname), ipfs_port)
```
vsbogd: This IPFS client code is copied three times as far as I see. Could you please wrap it to separate function and reuse.

snet_cli/commands.py
arturgontijo: Even if the endpoint is unreachable the user can type "y" and proceed with the publishing/update.
If we check it with something like `urlparse` (is this what you're suggesting?) some mistyping can pass unnoticed, like: http://singulritynet.io:7000.

snet_cli/commands.py
arturgontijo: I'll create a function to use this `protoc` part of code.

snet_cli/commands.py
arturgontijo: Sure!

snet_cli/commands.py
vsbogd: Ah ok didn't catch it at first time. If we can proceed even if service is down that is ok I think. Thanks!

vsbogd:
Looks good, thanks!

Issue #69, opened 2018-10-03
bip32utils dependency is not satisfied


stellarspot:
setup.py contains dependency 'bip32utils==0.3.post3'
Now it does not work:
```
> sudo pip3 install bip32utils==0.3.post3
Collecting bip32utils==0.3.post3
  Could not find a version that satisfies the requirement bip32utils==0.3.post3 (from versions: )
No matching distribution found for bip32utils==0.3.post3

> sudo pip3 install bip32utils
Collecting bip32utils
  Could not find a version that satisfies the requirement bip32utils (from versions: )
No matching distribution found for bip32utils
```
It seems that the git repository that contains Python api for BIP32 HD key nodes is removed:
https://github.com/prusnak/bip32utils

stellarspot:
The python3-bip32utils is still present in debian packages:
https://packages.debian.org/sid/python3-bip32utils

arturgontijo:
I'm checking this.

vsbogd:
Thanks!

stellarspot:
https://github.com/prusnak/bip32utils is the fork of 
https://github.com/lyndsysimon/bip32utils
with some fixes

arturgontijo:
The package itself is still at PyPI at [bip32utils-0.3.post3.tar.gz](https://files.pythonhosted.org/packages/53/43/5be31c3ca597334bb8626eee3c19a5e494d198d128d1021ed6151d5c9ab2/bip32utils-0.3.post3.tar.gz).
I'm trying to figure out what happened with prusnak repo, why he removed it.

arturgontijo:
From prusnak:
"It was unmaintained code, so I deleted it.
I recommend using https://github.com/richardkiss/pycoin or
https://github.com/ranaroussi/pywallet for BIP32"

Checking/adapting suggestions now...

PR #71, opened 2018-10-03
[bip32] Using pycoin

arturgontijo:
Related #69
Now we'll use [pycoin](https://github.com/richardkiss/pycoin) to handle HD wallets (BIP32).

### DIFF
--- a/snet_cli/identity.py
+++ b/snet_cli/identity.py
@@ -16,6 +16,8 @@ from trezorlib.transport_hid import HidTransport
 from snet_cli._vendor.ledgerblue.comm import getDongle
 from snet_cli._vendor.ledgerblue.commException import CommException
 
+BIP32_HARDEN = 0x80000000
+
 
 class IdentityProvider(abc.ABC):
     @abc.abstractmethod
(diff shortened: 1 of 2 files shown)


snet_cli/identity.py
```diff
@@ -117,18 +117,18 @@ def __init__(self, w3, index):
         self.client = TrezorClient(HidTransport.enumerate()[0])
         self.index = index
         self.address = self.w3.toChecksumAddress(
-            "0x" + bytes(self.client.ethereum_get_address([44 + bip32utils.BIP32_HARDEN,
-                                                           60 + bip32utils.BIP32_HARDEN,
-                                                           bip32utils.BIP32_HARDEN, 0,
+            "0x" + bytes(self.client.ethereum_get_address([44 + 0x80000000,
```
vsbogd: I would propose introduce the constant for this number.

vsbogd:
Looks good for me besides lack of constant for BIP32 hardened key indices base.

snet_cli/identity.py
arturgontijo: My first thought was to use it from `pycoin` but it seems to use the hex "directly" too.
But I agree with you, changing...

PR #77, opened 2018-10-18
add print_my_channels option for mpe-client

astroseger:
Add print_my_channels option to mpe-client (print all channels which belongs to the client). This is the first element in stateless client logic.  
Moreover I've added helper function to class BlockchainCommand: call_contract_command and transact_contract_command. Using this functions we can remove all ugly "ContractCommand" calls.

### DIFF
--- a/snet_cli/utils.py
+++ b/snet_cli/utils.py
@@ -183,3 +183,13 @@ def compile_proto(entry_path, codegen_dir, proto_file=None):
 
     except Exception as e:
         return False
+
+# return element of abi (return None if fails to find)
+def abi_get_element_by_name(abi, name):
+    for a in abi["abi"]:
+        if ("name" in a and a["name"] == name):
+            return a
+    return None
+
+def abi_decode_struct_to_dict(abi, struct_list):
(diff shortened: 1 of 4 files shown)


snet_cli/utils.py
```diff
@@ -183,3 +183,13 @@ def compile_proto(entry_path, codegen_dir, proto_file=None):
 
     except Exception as e:
         return False
+
+# return element of abi (return None if fails to find)
+def abi_get_element_by_name(abi, name):
+    for a in abi["abi"]:
```
arturgontijo: Will "abi" key always be within abi dict?
If not or abi == None, this for loop will break.
I'd suggest use something like `if abi and "abi" in abi:`
Offtopic: There are too much off `abi` in this line...

arturgontijo:
> Add print_my_channels option to mpe-client (print all channels which belongs to the client). This is the first element in stateless client logic.
> Moreover I've added helper function to class BlockchainCommand: call_contract_command and transact_contract_command. Using this functions we can remove all ugly "ContractCommand" calls.

The call/transact_contract_command is a very good improvement!

snet_cli/utils.py
astroseger: done :)

PR #85, opened 2018-10-25
add basic mpe-service functionality for publish service

astroseger:
add basic mpe-service functionality for dealing with service metadata and for publishing services in Registry (in the context of MPE payment system).

### DIFF
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -517,7 +517,7 @@ def add_mpe_client_options(parser):
     add_p_channel_id(p)
     
     # "call_server":  call server using the payment channel in stateless manner (protobuf should be already compiled)
-    p = subparsers.add_parser("call_server", help="call server in stateless manner. We ask state of the channel from the server.")
+    p = subparsers.add_parser("call_server", help="call server in stateless manner. We ask state of the channel from the server. Protobuf should be already compiled.")
     p.set_defaults(fn="call_server_statelessly")
     add_p_mpe_address(p)
     add_p_channel_id(p)
@@ -525,7 +525,7 @@ def add_mpe_client_options(parser):
     add_p_full_service_for_call(p)                                                
     
     # "call_server_lowlevel":  low level function for calling the server using already compiled protobuf
-    p = subparsers.add_parser("call_server_lowlevel", help="Low level function for calling the server")
+    p = subparsers.add_parser("call_server_lowlevel", help="Low level function for calling the server. Protobuf should be already compiled")
     p.set_defaults(fn="call_server_lowlevel")
     add_p_full_message(p)
     add_p_full_service_for_call(p)
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -9,7 +9,8 @@ from snet_cli.commands import IdentityCommand, SessionCommand, NetworkCommand, C
 from snet_cli.identity import get_identity_types
 from snet_cli.session import get_session_keys
 from snet_cli.utils import type_converter, get_contract_def
-from snet_cli.mpe_client_command import MPEClientCommand
+from snet_cli.mpe_client_command  import MPEClientCommand
+from snet_cli.mpe_service_command import MPEServiceCommand
 
 
 class CustomParser(argparse.ArgumentParser):
@@ -85,8 +86,11 @@ def add_root_options(parser, config):
     organization_p = subparsers.add_parser("organization", help="Interact with SingularityNET Organizations")
     add_organization_options(organization_p)
 
-    mpe_client_p = subparsers.add_parser("mpe-client", help="Interact with SingularityNET services (using MultiPartyEscrow channels)")
+    mpe_client_p = subparsers.add_parser("mpe-client", help="Interact with SingularityNET services in the context of MPE payment system")
     add_mpe_client_options(mpe_client_p)
+
+    mpe_server_p = subparsers.add_parser("mpe-service", help="Functionality for publish service in the context of MPE payment system")
+    add_mpe_service_options(mpe_server_p)
 
 
 def add_version_options(parser):
@@ -505,14 +509,14 @@ def add_mpe_client_options(parser):
         p.add_argument("amount",     type=int, help="amount")
 
 
-    # "complie_from_file": Compile protobuf from the file. We will use it for the given channel (channel_id)
+    # "compile_from_file": Compile protobuf from the file. We will use it for the given channel (channel_id)
     p = subparsers.add_parser("compile_from_file", help="Compile protobuf from the file")
     p.set_defaults(fn="compile_protobuf_from_file")
     p.add_argument("proto_dir",  type=str, help="protobuf .proto directory")
     p.add_argument("proto_file", type=str, help="protobuf .proto file")
     add_p_channel_id(p)
     
-    # "call_server":  call server using the payment channel
+    # "call_server":  call server using the payment channel in stateless manner (protobuf should be already compiled)
     p = subparsers.add_parser("call_server", help="call server in stateless manner. We ask state of the channel from the server.")
     p.set_defaults(fn="call_server_statelessly")
     add_p_mpe_address(p)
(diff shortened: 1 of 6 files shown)


snet_cli/mpe_service_command.py
```diff
@@ -0,0 +1,101 @@
...
+    # publis proto files in ipfs and print hash
+    def publish_proto_in_ipfs(self):
+        ipfs_hash_base58 = utils_ipfs.publish_proto_in_ipfs(self._get_ipfs_client(), self.args.protodir)
+        self._printout(ipfs_hash_base58)
+        
+    # Init metadata and set model_ipfs_hash  
+    def metadata_init(self):
+        metadata = mpe_service_metadata()
+        metadata.set_model_ipfs_hash(self.args.model_ipfs_hash)
+        metadata.set_mpe_address(self.args.mpe_address)
```
vsbogd: Side note on keeping MultiPartyEscrow (MPE) contract address in metadata.

It is required to correctly handle MPE contracts updates in both client and service sides. MPE contract address is used in signed payment message to guarantee that both payment sender and payment receiver use the same data source to check the payment. 

When data is kept in contract then contract address changes during contract upgrade and service provider should publish new Service with new metadata to incorporate change.

When data and logic of contract are decoupled the upgrade may be seamless for sender and receiver as  most of upgrades doesn't affect data contract address. And changing this address in metadata is required only when data part of the contract is updated.

snet_cli/mpe_service_command.py
```diff
@@ -0,0 +1,101 @@
+from snet_cli.commands    import BlockchainCommand
+import snet_cli.utils_ipfs as utils_ipfs
+from snet_cli.mpe_service_metadata import mpe_service_metadata, load_mpe_service_metadata
+from snet_cli.utils import type_converter
+import base58
```
vsbogd: Should be added as dependency to the ```setup.py```

snet_cli/utils_ipfs.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# make tar from protodir/*proto, and publish this tar in ipfs
+# return base58 encoded ipfs hash
+def publish_proto_in_ipfs(ipfs_client, protodir):
+    files = glob.glob(os.path.join(protodir, "*.proto"))
+    
+    # We are sorting files before we add them to the .tar since an archive containing the same files in a different
+    # order will produce a different content hash;
+    files.sort()
```
vsbogd: May be I missed something, but it seems that ```files``` collection contains full paths which in turn has prefix passed by user in command line. So it would be more correct to sort only basenames as only they are packed into tarball.

snet_cli/utils_ipfs.py
```diff
@@ -0,0 +1,30 @@
...
+    for f in files:
+        tar.add(f, os.path.basename(f))
+    tar.close()
+    return ipfs_client.add_bytes(tarbytes.getvalue())
+
+# get file from ipfs
+# We must check the hash becasue we cannot believe that ipfs_client wasn't been compromise
+def get_from_ipfs_and_checkhash(self, ipfs_client, ipfs_hash_base58):
+    data  = ipfs_client.cat(ipfs_hash_base58)
+    print("!!! We must check that hash in IPFS is correct (we cannot be sure that ipfs is not compromized) !!! Please implement it !!!")
```
vsbogd: I would propose adding an issue on this as well and explain design decision there.

snet_cli/mpe_service_metadata.py
```diff
@@ -0,0 +1,101 @@
+# functions for manipulate with server.json file
+# metadata format:
+#----------------------------------------------------
+# version         - used to track format changes (current version is 1)
+# model_ipfs_hash - IPFS HASH to the .tar archive of protobuf service specification
+# mpe_address     - Address of MultiPartyEscrow contract. 
+#                   Client should use it exclusively for cross-checking of mpe_address, 
+#                        (because service can attack via mpe_address)
+#                   Daemon can use it directly if authenticity of metadata is confirmed
```
vsbogd: Side note on keeping MultiPartyEscrow (MPE) contract address in metadata.

MPE contract address is also required to correctly handle MPE contracts updates in both client and service sides. MPE contract address is used in signed payment message to guarantee that both payment sender and payment receiver use the same data source to check the payment.

When data is kept in contract then contract address changes during contract upgrade and service provider should publish new Service with new metadata to incorporate change.

When data and logic of contract are decoupled the upgrade may be seamless for sender and receiver as most of upgrades doesn't affect data contract address. And changing this address in metadata is required only when data part of the contract is updated.

snet_cli/mpe_service_metadata.py
```diff
@@ -0,0 +1,101 @@
...
+# pricing {}      - Pricing model
+#          Possible pricing models:
+#          1. Fixed price
+#              price_model  - "fixed_price"
+#              price        -  unique fixed price for all method
+#              (other pricing models can be easily supported)
+# groups[]       - group is the number of endpoints which shares same payment channel; 
+#                   grouping strategy is defined by service provider; 
+#                   for example service provider can use region name as group id
```
vsbogd:
```suggestion
#                   for example service provider can use region name as group name
```

snet_cli/mpe_service_metadata.py
```diff
@@ -0,0 +1,101 @@
...
+                              "group_id"        : str(group_id_base64),
+                              "payment_address" : payment_address}]
+        return group_id_base64
+    
+    def add_endpoint(self, group_name, endpoint):
+        if (not self._is_group_present(group_name)):
+            raise Exception("the group %s is not present"%str(group_name))
+        e = {"group_name" : group_name, "endpoint"   : endpoint}
+        if (e in self.m["endpoints"]):
+            raise Exception("We alread have endpoint %s in group %s"%(endpoint, group_name))
```
vsbogd:
```suggestion
            raise Exception("We already have endpoint %s in group %s"%(endpoint, group_name))
```

snet_cli/mpe_service_metadata.py
```diff
@@ -0,0 +1,101 @@
...
+        for g in groups:
+            if (g["group_name"] == group_name):
+                return True
+        return False
+
+    def get_json(self):
+        return json.dumps(self.m)
+    
+    def set_from_json(self, j):
+        # TODO: we should consistensy here
```
vsbogd: Could you please explain this comment - I did't get it.

snet_cli/arguments.py
```diff
@@ -554,3 +558,103 @@ def add_p_full_message(p):
...
+    add_p_service_path_opt(p)
+    add_p_metadata_file_opt(p)
+    add_p_tags_opt(p)
+    add_p_transact_yes(p)
+
+    p = subparsers.add_parser("get_service_metadata", help="Get service metadata from registry")
+    p.set_defaults(fn="get_service_metadata_hash_from_registry")
+    add_p_registry_info(p)
+
+    p = subparsers.add_parser("publish_service_fixed_price_single_group", help="Publish service with fixed price, single group and multiply endpoints")
```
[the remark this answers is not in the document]
astroseger: Yes, I agree. It is a high level function which do everything at once, and you could do the same make several low level calls.

- publish_proto
- metadata_init
- metadata_add_group group1
- metadata_add_endpoints
- publish_service

Maybe we could remove this high level function at all... I will keep it for now...

snet_cli/utils_ipfs.py
astroseger: All files are from the same directory. I don't support protobuf for several directories for now.

snet_cli/arguments.py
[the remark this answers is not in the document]
vsbogd: It is not a blocker, please interpret it as an opinion and an idea.

snet_cli/mpe_service_metadata.py
astroseger: Yes. I've changed it also here https://github.com/singnet/wiki/blob/master/multiPartyEscrowContract/MPEServiceMetadata.md

snet_cli/mpe_service_metadata.py
astroseger: we should discuss it here https://github.com/singnet/wiki/blob/master/multiPartyEscrowContract/MPEServiceMetadata.md

Issue #93, opened 2018-10-28
members list should be optional in 'snet organization create'


ferrouswheel:
To create a personal org, I need to do:

> snet organization create joel MY_ADDRESS

But snet already has my identity, so I wonder if it should be sufficient to just do:

> snet organization create joel

I'm also unclear who the owner of the organisation is if I specify a different address. Since I'm making the transaction, am I automatically the owner? Or is that role granted upon the initial members?

If I'm automatically the owner, then I think having an empty member list should be possible.

arturgontijo:
Yeah, the owner is the user that creates the organization.
Looking at [Registry.sol](https://github.com/singnet/platform-contracts/blob/master/contracts/Registry.sol#L127) I saw that's possible to create an organization without any "extra" members (with only the owner).
I'll fix this.

Issue #86, opened 2018-10-26
Error if you provide empty tags to service init


ferrouswheel:
```
Input a list of tags for your service, space separated: (default: [])

Error: 'list' object has no attribute'read'
```

I assume the the default value is being provided as a list, and the parsing code is attempting to read this.

Issue #27, opened 2018-08-10
Make `snet service init` use float argument for AGI price


erickearns:
Prompt currently claims it expects the price in AGI, but it appears to actually be in `10 ** -8` AGI.

ferrouswheel:
Exact prompt is `Choose a price in AGI to call your service: (default: 0)`.

Agree that this is confusing and needs to either auto-convert to the AGI minimum unit, or be clear it's in units of `10^-8`.

Should also have a sensible default instead of 0. I think 1 AGI should be the default i.e. `1x10^8` of the minimum divisible unit.

astroseger:
We have the term for 10^-8 AGI : 'cogs'  (see AGI token contract). So I propose to replace 
```bash
Choose a price in AGI to call your service: (default: 0)
```
with
```bash
Choose a price in cogs (cogs = 10^-8 AGI) to call your service: (default: 0)
```
I'm writing the code for publishing the service in MPE payment system. And I was going to use the term cogs.

PR #95, opened 2018-10-29
#93, #86 and #27

arturgontijo:
Related: #93, #86, #27
`snet organization create` without members (owner-only).
Accept empty tag list and price in cogs at `service init`.


arturgontijo:
@stellarspot, @vsbogd and @astroseger could you please check if some os these changes might break the pipeline? I checked but you guys are more familiar with it than me.

Issue #117, opened 2018-11-21
Update README.md with MPE changes


raamb:
Update the examples in the README.md to use MPE

PR #130, opened 2018-11-26
Upated README with MPE changes

raamb:
#117 Updated the commands described in the README to use MPE. Also cleaned up references to the agent contracts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ snet client open_init_channel_registry [-h] [--registry REGISTRY]
            organization service amount
            expiration
 ```
-* Open channel with our specified service
+* Open channel with the specified service
  * `organization`:name of organization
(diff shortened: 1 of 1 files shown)


astroseger:
The best solution would be to autogenerate part of this readme from snet-cli command line help. Probably there are already some tools for doing this (for example this one https://sphinx-argparse.readthedocs.io/en/stable/).... But probably not for this time :)

raamb:
> The best solution would be to autogenerate part of this readme from snet-cli command line help. Probably there are already some tools for doing this (for example this one https://sphinx-argparse.readthedocs.io/en/stable/).... But probably not for this time :)

Agree we should do that. For now if the key commands are covered correctly, please approve, I will merge this.

PR #135, opened 2018-11-29
Specifying Python version in scripts/blockchain

ramongduraes:
When trying to run "scripts/blockchain" from ubuntu having both Python3.5 and Python3.6 installed, Python3.5 is chosen by default. That results in a very generic error being returned. Specifying the Python version fixes the problem!

### DIFF
--- a/scripts/blockchain
+++ b/scripts/blockchain
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python3.6
 
 import os
 import pathlib


astroseger:
This solution is not very generic. We will not be able to run it with  python3.7.

astroseger:
By the way, we could use this https://www.scivision.co/require-minimum-python-version-setup-py/ in setup.py

pennachin:
Agreed, we shouldn't specify python minor versions in the scripts themselves, this is a user configuration issue (perhaps Ramon's ubuntu is defaulting to 3.5 when it apparently shouldn't?)

ramongduraes:
Yeah, agreed! It has happened to Artur, Glauter and André as well so maybe it's something to keep in mind.

Issue #119, opened 2018-11-21
snet organization create does not print error message


stellarspot:
I started IPFS, migrated contracts and run ganache-cli.
When I call 'snet organization create' it prints nothing as it passed, but it really fails without any error message:

```
snet organization create TestOrg  -y 
echo $?
1
```

astroseger:
@stellarspot  could you check it again in your environment? It might be solved in #136

astroseger:
closed with #142

Issue #73, opened 2018-10-04
"Web3.py only accepts checksum addresses...Or, if you must accept lower safety, use Web3.toChecksumAddress(lower_case_address)."


arturgontijo:
To reproduce:
```
$ snet set current_agent_at 0x1e89d9ed5bcc22f934af631cda771019081e57b2
set current_agent_at 0x1e89d9ed5bcc22f934af631cda771019081e57b2
```
Then running `snet client call [method] [json]`, I got this:
```
...
File "/usr/local/lib/python3.6/dist-packages/web3/utils/validation.py", line 123, in validate_address
    value,
web3.exceptions.InvalidAddress: ('Web3.py only accepts checksum addresses. The software that gave you this non-checksum address should be considered unsafe, please file it as a bug on their platform. Try using an ENS name instead. Or, if you must accept lower safety, use Web3.toChecksumAddress(lower_case_address).', '0x1e89d9ed5bcc22f934af631cda771019081e57b2')
```
But getting the address from etherscan (`0x1E89D9ed5bCC22F934AF631CdA771019081E57B2`), it works.

Question is...should we use this `lower safety` function or just warn user to change address format?

astroseger:
I would propose to do following:
for contract address we should accept lower case address because ganache, at the moment, print contract address in lower case. 
for account address we should accept only checksum address.

arturgontijo:
I think the best way will be always call `type_converter("address")(INPUT_ADDRESS)` before make a transaction that uses any `INPUT_ADDRESS`.
Maybe you're already doing this with MPE commands.

stellarspot:
I run the command
```
snet service metadata_init service/service_spec DNNModelService 0x3b2b3c2e2e7c93db335e69d827f3cc4bc2a2a2cb --multipartyescrow 0x5c7a4290f6f8ff64c69eeffdfafc8644a4ec3a4e
```

where the 0x3b2b3c2e2e7c93db335e69d827f3cc4bc2a2a2cb is the second address from the Ganache output and it fails with error:
```
Error: payment_address parameter is not a valid Ethereum checksum address
```
If I use the address from the [test](https://github.com/astroseger/my-platform-tests/blob/master/test1/run_service.sh) 0x3b2b3C2e2E7C93db335E69D827F3CC4bC2A2A2cB  it passes.

Is it intentional behavior?

astroseger:
yes... it is expected behavior. User should use only chacksum addresses. At least Ethereum actively promote it.

[...]

PR #142, opened 2018-12-05
Accounts only in checksum

astroseger:
This PR makes following: 
- We check that all Ethereum addresses, are in checksum format. There is only one exception. We allow contract addresses to be in not checksum format (#73 )
- Now "snet organization ..." will raise exception (not simply return without error code) in different abnormal  cases (#119 )
- We catch all exception in main (not only inside "getattr(args.cmd(conf, args), args.fn)()")
- add tests for "snet organization ...."
- add tests for "snet contract ...."

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -16,6 +16,7 @@ from snet_cli.config import get_session_identity_keys, get_session_network_keys
 from snet_cli.utils_config import get_contract_address, get_field_from_args_or_session
 from snet_cli.identity import RpcIdentityProvider, MnemonicIdentityProvider, TrezorIdentityProvider, \
     LedgerIdentityProvider, KeyIdentityProvider
+import web3
 
 
 class Command(object):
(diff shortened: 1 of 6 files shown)


snet_cli/commands.py
```diff
@@ -355,18 +361,18 @@ def change_owner(self):
...
-        new_owner = new_owner if new_owner.startswith("0x") else "0x" + new_owner
-        if new_owner == owner:
-            self._printerr("\n{} is the owner of!\n".format(self.args.owner, self.args.name))
-            return
+        new_owner = self.args.owner
+        if not web3.eth.is_checksum_address(new_owner):
+            raise Exception("New owner account %s is not a valid Ethereum checksum address"%new_owner)
+
+        if new_owner.lower() == owner.lower():
+            raise Exception("\n{} is the owner of {}!\n".format(new_owner, self.args.name))
```
Qodestackr, line 371: How will I handle this from infura?

arturgontijo: @Qodestackr we need to be sure that the address is in a valid `checksum` format, ie:
```
>>> public_key
'0x20802d1a9581b94e51db358c09e0818d6bd071b4'
>>> w3.toChecksumAddress(public_key)
'0x20802d1a9581b94e51db358C09e0818d6bd071b4'
```
Notice that some hex are in uppercase when in a valid checksum format.

PR #145, opened 2018-12-08
reverted workaround introduced in PR #59

Chris-Ricketts:


setup.py
```diff
@@ -16,7 +16,7 @@
     install_requires=[
         'grpcio-tools==1.14.1',
         'jsonrpcclient==2.5.2',
-        'eth-typing<2', # workaround until next version after eth-abi v2.0.0-beta.1 is released
+        'eth-typing>=2', # workaround until next version after eth-abi v2.0.0-beta.1 is released
```
astroseger: comments can be removed...

Chris-Ricketts: Done - don't know how I missed that.

setup.py
```diff
@@ -16,7 +16,7 @@
     install_requires=[
         'grpcio-tools==1.14.1',
         'jsonrpcclient==2.5.2',
-        'eth-typing<2', # workaround until next version after eth-abi v2.0.0-beta.1 is released
+        'eth-typing>=2',
```
stellarspot: This line was added because of the issue #34 Fix issue with pycryptodome dependencies rised by new eth-hash v0.2.0

May be it is better just to remove this line at all?
Will eth-typing version 2 be got as dependency resolution in this case?

astroseger:
@Chris-Ricketts Please remove this line completely, and we will merge it. Thanks!

Chris-Ricketts:
Done - thanks for your patience!

Issue #128, opened 2018-11-23
Add ability to set channel signer when channel is opened


vsbogd:
Add command line option to set channel signer which is not equal to current identity when channel is being opened.

Typical usage scenario:
- switch ```snet-cli``` to the channel sender identity
- open channel with specific signer identity
- switch ```snet-cli``` to use signer identity
- make a service call

astroseger:
was solved in #148

PR #148, opened 2018-12-13
Fix for platform contracts v0.2.5

### LINKED ISSUE #143: Add --force to snet-cli create network
At the moment ```snet-cli create network``` fails if Ethereum endpoint is not up and running. It would be useful to have a way to skip this check as ```snet-cli``` configuration could be wirtten properly even if Ethereum network is not available at the moment. I would suggest adding ```--force``` or ```--skip-check``` property to do this.

astroseger:
This PR does following:
* adapt snet-cli for new platform-contracts (https://github.com/singnet/platform-contracts/pull/84 and https://github.com/singnet/platform-contracts/pull/84)
     * In ```snet create organization``` by default user should specify only org_name. In this case org_id will be generated automatically. 
     * org_id is specified in string format (similar to old organization name in previous version)
     * User can specify org_id by hands: ```snet create organization myOrgName --org-id myOrgId```
     * service_name was renamed to service_id (but there is not API change here)
     * In all function (except ```snet organization create```) user should specify org_id and service_id (not org_name and service_display_name).
     * org_id is printed in the last line of ```snet create organization```, so user can easily save it and use in the following functions.
```
ORG_ID=`snet organization create myOrgName -y -q | tail -1`
# now we can use $ORG_ID
snet organization add-members $ORG_ID 0x32267d505B1901236508DcDa64C1D0d5B9DF639a -y
snet service publish $ORG_ID MyServiceID 
```

* Add --skip-check in ```snet network create``` (fix #143)
* Add CI for ```snet network```

CI will be green for new release of platform-contracts.

### DIFF
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -7,9 +7,10 @@ from pathlib import Path
 from snet_cli.commands import IdentityCommand, SessionCommand, NetworkCommand, ContractCommand, OrganizationCommand, VersionCommand
 from snet_cli.identity import get_identity_types
 from snet_cli.utils import type_converter, get_contract_def
-from snet_cli.mpe_client_command  import MPEClientCommand
-from snet_cli.mpe_account_command  import MPEAccountCommand
+from snet_cli.mpe_account_command import MPEAccountCommand
 from snet_cli.mpe_service_command import MPEServiceCommand
+from snet_cli.mpe_channel_command import MPEChannelCommand
+from snet_cli.mpe_client_command  import MPEClientCommand
 from snet_cli.utils_agi2cogs import stragi2cogs
 from snet_cli.config import get_session_keys, get_session_network_keys_removable
 
@@ -75,6 +76,10 @@ def add_root_options(parser, config):
 
     p = subparsers.add_parser("account", help="AGI account")
     add_mpe_account_options(p)
+
+    p = subparsers.add_parser("channel", help="Interact with SingularityNET payment channels")
+    add_mpe_channel_options(p)
+
     mpe_client_p = subparsers.add_parser("client", help="Interact with SingularityNET services")
     add_mpe_client_options(mpe_client_p)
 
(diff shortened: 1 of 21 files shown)


snet_cli/commands.py
```diff
@@ -10,13 +10,16 @@
 
 from snet_cli.contract import Contract
 from snet_cli.identity import get_kws_for_identity_type, get_identity_types
```
arturgontijo: `get_identity_types` is unused.
`from snet_cli.config import ...` too.

test/functional_tests/script1_twogroups.sh
```diff
@@ -19,7 +19,7 @@ ipfs cat $IPFS_HASH > service_metadata2.json
 # compare service_metadata.json and service_metadata2.json
 cmp <(jq -S . service_metadata.json) <(jq -S . service_metadata2.json)
 
-snet organization create testo -y -q
+snet organization create org1 --org-id testo -y -q
```
arturgontijo: Is `org1` here correct?
In the other files you user `... testo --org-id testo ...`

astroseger: yes org1 is just organization name. Later we always refer to organization via org_id.

raamb:
@astroseger the org_id is important for the users to interact with the snet-cli (and to seed in the daemon config) so my preference is to not auto generate it by default. 

My proposal is 
1. User to provide an orgId in `snet create organization`
2. If user does not provide orgId error out
3. Provide an option to  `snet create organization` command to auto generate org id.

This way org_id is auto generated when the user requests for it and our default behaviour is to ask for an org_id.

astroseger:
After discussion with @raamb  the following modifications have been made:
* By default user should specify ```--org-id``` in ```snet organization create```. User can still auto-generate org-id by specifying ```--auto``` options (actually --org-id and --auto are mutually exclusive options so user should specify one of them).
```
# specify org-id by hands (expected behavior )
snet organization create myOrgName --org-id myOrgId -y -q
# or auto-generate org-id
snet organization create myOrgName --auto -y -q
```

* ```snet organization list-my``` was added. It prints organization which has the current identity as the owner or as a member

* ```snet organization list-org-names``` print list of organization with their names (this function was initially presented in this PR, but I forgot to mention it in my first comment)

astroseger:
@raamb @arturgontijo @vsbogd  please review new changes. 
I've add the following changes in this PR. 

* Fix security issue with tar files (using tarfile.extractall without prior inspection is a serious security hazard!) 


* use '-'  in all parameters #155. There is a lot of API change. In #155 I posted  a script for automatically update your shell scripts. 

*  I split ```snet client```  into three sections:
      * ```account``` -- manipulation with AGI account. I moved balance, deposit, withdraw and transfer here.
      *  ```channel``` -- manipulation with payment channels.
      *  ```client``` --- Interact with SingularityNET services. Only call, call-lowlevel, get-channel-state stays here.

* Solution for #151 was added. Old ```print_initialized_channels``` and ```print_all_channels``` was replaced by 6 different functions for print and search for channels. Please find more info in my last comment in #151, I will not duplicate it here.

* add ```snet service get-api-metadata``` and ```snet service get-api-registry``` (issue #149 )

* add ```--endpoints``` and ```--fixed-price``` for ```snet service metadata-init``` (see issue #152). Now if you have single payment group you can create your metadata in one shot:
```
snet service metadata-init <proto_dir> <service_display_name> <payment_address> --fixed-price <fixed_price> --endpoints <your endpoints>
```

* I've added option to set channel signer  when channel is open (#128 )
 
* Fix small issue with ```snet channel claim-timeout```. Now we check for empty channel before sending transaction (and we do not send transaction if channel is already empty).

snet_cli/arguments.py
```diff
@@ -393,35 +389,71 @@ def add_p_open_channel_basic(p):
...
+    add_p_metadata_file_opt(p)
+
+    p = subparsers.add_parser("claim-timeout", help="Claim timeout of the channel")
     p.set_defaults(fn="channel_claim_timeout")
     add_p_channel_id(p)
     add_p_mpe_address_opt(p)
     add_transaction_arguments(p)
 
-    p = subparsers.add_parser("channel_extend_add", help="Set new exporation for the channel and add funds")
+    p = subparsers.add_parser("extend-add", help="Set new exporation for the channel and add funds")
```
raamb: Minor - Help should "Set new expiration for the channel and add funds"

snet_cli/commands.py
```diff
@@ -277,10 +283,13 @@ def transact(self):
 
 
 class OrganizationCommand(BlockchainCommand):
-    def _getorganizationbyname(self):
-        return self.call_contract_command("Registry", "getOrganizationByName", [type_converter("bytes32")(self.args.name)])
+    def _getorganizationbyid(self, org_id):
+        org_id_bytes32 = type_converter("bytes32")(org_id)
+        if (len(org_id_bytes32) > 32):
+            raise Exception("Your org_id is too long: len(org_id_bytes32)=%i"%(len(org_id_bytes32)))
```
raamb: Lets include the allowed length of 32 bytes in the message something like below.
`raise Exception("Your org_id is too long, it should be 32 bytes or less. len(org_id_bytes32)=%i"%(len(org_id_bytes32)))`

snet_cli/commands.py
```diff
@@ -294,137 +303,186 @@ def get_members_from_args(self):
...
+        org_id = self.args.org_id
         # Check if Organization exists and member is not part of it
-        (found, _, _, members, _, _) = self._getorganizationbyname()
-        if not found:
-            raise Exception("\n{} doesn't exist!\n".format(self.args.name))
+        (found, _, _, _, members, _, _) = self._getorganizationbyid(org_id)
+        self.error_organization_not_found(org_id, found)
 
         members = [member.lower() for member in members]
+        print(members)
```
raamb: Do we need to print members here?

Issue #150, opened 2018-12-15
Add service description field into service metadata


vsbogd:
It's worth to add some service description field to the service metadata. It could contain the URL which points to the service API documentation. It will allow users to be familiar with API before using it. This issue is tightly coupled with #149

astroseger:
was solved  by https://github.com/singnet/snet-cli/pull/160

PR #160, opened 2018-12-25
Add snet service_description field in metadata

astroseger:
This PR solves #150.

It adds "snet service metadata-add-description-json" to add service_description field in metadata. Description should be passed in json format. For example: 
```
snet service metadata-add-description-json '{"description_string":"string1","description_int":1,"description_dict":{"a":1,"b":"s"}}'
```

### DIFF
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -595,6 +595,11 @@ def add_mpe_service_options(parser):
     p.add_argument("--group-name", default=None, help="name of the payment group to which we want to add endpoints. Parameter should be specified in case of several payment groups")
     add_p_metadata_file_opt(p)
 
+    p = subparsers.add_parser("metadata-add-description-json", help="Add service description")
+    p.set_defaults(fn="metadata_add_description_json")
+    p.add_argument("json", default=None, help="Service description in json")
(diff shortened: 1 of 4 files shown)


astroseger:
The question here: should we make "description" and "uri" fields obligatory?

astroseger:
I could replays ```snet service metadata-add-description-json``` with ```snet service metadata-add-description``` with 3 optional parameters:
* ```--description```
* ```--uri```
* ```--json```  (all other fields in json format)

If ```--description``` and ```--json``` both present then json should not contain "description" field, and the same for ```--uri```.

astroseger:
So... What we are doing with this PR? Should I implement what I've proposed?

arturgontijo:
I think this is very important, the URI could point to User's Guide and this is fantastic!
@astroseger, I like your 3 optional parameters idea.

astroseger:
Should I use "uri" or "url"?  I've implemented my proposition with  ```snet service metadata-add-description``` and 3 optional parameters (```--json```, ```---description```, ```--uri```)

arturgontijo:
Thinking again I suggest using URL.

Issue #134, opened 2018-11-29
Add command to allow service provider claiming funds from channel


vsbogd:
At the moment channel claiming is implemented within ```snet-daemon``` through ```snet-daemon claim``` interface. This command should be moved into the ```snet-cli``` tool.

Possible way to implement it is to add additional gRPC interface to the daemon to:
- claim payments
- list payments in progress
- list channels

This interface will be implemented by separate daemon instance: treasurer server.

Use this interface to implement ```snet-cli claim``` command.

See https://github.com/singnet/snet-cli/issues/134#issuecomment-443182697 for more detailed description

astroseger:
There is a problem. 

Which ethereum identity will be used? snet-cli identity or identity of this instance of the daemon?

1. I propose to agree in terminology. I propose to call "the treasurer server" the system which actually sign transaction on behalf of service provider.

Now I can ask the main question.
Who will be the treasurer server (who will sign the transactions): snet-cli or the instance of the daemon? 

Because it will be the daemon, it kind of break all logic of snet-cli as owner of identity.

astroseger:
I would argue that in the beta, it would be better to make snet-cli the treasurer server (only snet-cli will actually send transactions).
It will allow as to solve the following issues in one place: 
- https://github.com/singnet/snet-cli/issues/111 (poor usability on the slow main net)
- https://github.com/singnet/snet-cli/issues/67  (algorithm of sending transaction which will actually work on the main net)

Because if we send transaction in snet-cli **and** in the daemon we need to solve these issue in two places.

vsbogd:
1. It will be code duplication: to write this server properly in snet-cli you will need to reproduce the code which is already implemented in daemon. 
2. Also supporting this infrastructure in daemon code base is simpler, because you have all things which plays with internal storage in one place.
3. Anyway one need clear interface for claiming and daemon already has it, the only work which should be done is to provide snet-cli access to this API.

https://github.com/singnet/snet-cli/issues/111 and https://github.com/singnet/snet-cli/issues/67 can be solved in daemon code base as well.

vsbogd:
Results of offline discussion with @astroseger:
- snet-daemon already has an API to work with payments (see [godoc](https://godoc.org/github.com/singnet/snet-daemon/escrow#PaymentChannelService))  such as: 
  - ```StartClaim()```
  - ```Claim.Finish()```
  - ```Claim.Payment()```
  - ```ListClaims()```
[...]

PR #162, opened 2018-12-27
Treasurer server

astroseger:
This PR add treasurer logic (solves #134). This PR also add 'dummy' daemon for testing of ```snet treasurer``` logic together with previously untested ```snet client call```.

We have the following functions:
- ```snet treasurer print-unclaimed``` - Print all unclaimed payments and total amount of unclaimed funds. 
- ```snet treasurer claim``` - Claim given channels. We also claim all pending 'payments in progress' in case we 'lost' some payments.
-  ```snet treasurer claim-all```  -    Claim all channels. We also claim all 'pending' payments.
-  ```snet treasurer claim-expired``` -  Claim all channels which are close to expiration date. This function have ```--expiration-threshold``` parameter: expiration threshold in blocks (default is 34560 ~ 6 days with 15s/block)

Service provide can simply call ```snet treasurer claim-expired``` with frequency depending on his payment_expiration_threshold defined in metadata. With default value of payment_expiration_threshold=40320 (one week) he could call ```claim-expired``` one time per day (it will be safe enough).
He can call ```snet treasurer claim-all``` when he need it.

There are two features which haven't been implemented in this PR. I will create issues for them.
1. In the current version we can specify only one  endpoint via parameter ```--endpoint```. We should get endpoints from metadata and use round robin.
2. I the current version we use MPE.channelClaim. But we should use MPE.multiChannelClaim in order to save time and gas.

### DIFF
--- a/snet_cli/mpe_treasurer_command.py
+++ b/snet_cli/mpe_treasurer_command.py
@@ -1,0 +1,158 @@
+from snet_cli.mpe_channel_command import MPEChannelCommand
+from snet_cli.utils import compile_proto
+from pathlib import Path
+import sys
+import os
+import grpc
+from eth_account.messages import defunct_hash_message
+from snet_cli.utils_proto import import_protobuf_from_dir
+from snet_cli.utils import int4bytes_big
+from snet_cli.utils_agi2cogs import cogs2stragi
+import web3
+
+
+# we inherit MPEChannelCommand because we need _get_channel_state_from_blockchain
+class MPETreasurerCommand(MPEChannelCommand):
+
+    def _sign_message_list_unclaimed(self, mpe_address, current_block):
+        message =  self.w3.soliditySha3(
+                        ["string",           "address",   "uint256"],
+                        ["__list_unclaimed", mpe_address, current_block])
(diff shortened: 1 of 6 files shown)


snet_cli/mpe_treasurer_command.py
```diff
@@ -0,0 +1,160 @@
...
+        proto_dir   = Path(__file__).absolute().parent.joinpath("resources", "proto")
+        if (not codegen_dir.joinpath("control_service_pb2.py").is_file()):
+            compile_proto(proto_dir, codegen_dir, proto_file = "control_service.proto")
+
+        stub_class, request_class, _ = import_protobuf_from_dir(codegen_dir, service_name)
+        return stub_class, request_class
+
+    def _decode_PaymentReply(self, p):
+        return {"channel_id" :  int4bytes_big(p.channel_id), "nonce" : int4bytes_big(p.channel_nonce), "amount" : int4bytes_big(p.signed_amount), "signature" : p.signature}
+
```
raamb: There is common code across methods with slight differences. We can consider writing common methods at a later point.

Issue #172, opened 2019-01-20
Validate endpoint is a URL


ferrouswheel:
Service endpoints don't get validated:

`snet service metadata-add-endpoints blah`
`snet service metadata-add-endpoints 'blah*&^%$#'`

Both the above are accepted and added to the metadata json.

This was raised in #49, then fixed in #60. Then later in #87 we suggest swapping to `urllib.parse` to avoid calling the url unnecessarily.

The MPE implementation doesn't do any checking.

astroseger:
It is related to https://github.com/singnet/snet-daemon/issues/153  and partly duplicate https://github.com/singnet/snet-cli/issues/127 . 
Let's discuss in  https://github.com/singnet/snet-daemon/issues/153 .

PR #179, opened 2019-01-20
Misc small changes

ferrouswheel:
- validates endpoint as url #172
- change top-commented functions and modules to use python docstrings instead
- clean up spacing on `print-channels` subcommand (header used double-spaces and one had a single-space, make everything use single-space so easy to process with awk etc)

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -457,8 +457,8 @@ class OrganizationCommand(BlockchainCommand):
             self._printerr("\nTransaction error!\nHINT: Check if you are the owner of {}\n".format(org_id))
             raise
 
-    # find organization with have the current identity as the owner or as the membmer
     def list_my(self):
+        """ Find organization that has the current identity as the owner or as the member """
         org_list = self.call_contract_command("Registry", "listOrganizations", [])
 
         rez_owner  = []
--- a/snet_cli/config.py
+++ b/snet_cli/config.py
@@ -141,12 +141,12 @@ class Config(ConfigParser):
         self._get_identity_section(identity)[key] = str(value)
         self._persist()
 
-    # return section for network or identity
     def _get_network_section(self, network):
+        """ return section for network or identity """
         return self[ "network.%s"%network ]
 
-    # return section for the specific identity
     def _get_identity_section(self, identity):
+        """ return section for the specific identity """
         return self[ "identity.%s"%identity ]
 
     def get_ipfs_endpoint(self):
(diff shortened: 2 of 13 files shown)


ferrouswheel:
Fails on this command `snet service metadata-add-endpoints 8.8.8.8:2020 9.8.9.8:8080 --group-name group1` - because both those endpoints don't specify the URL scheme (https/http).

I think we should either consistently require the scheme or never allow it.

astroseger:
1. see discussion here: https://github.com/singnet/snet-daemon/issues/153 ... We need to decide should we forbid ```<host>:<port>``` or not.  I would propose to translate it into ```http://<host>:<port>``` in MPEServiceMetadata.add_endpoint.

2.  I don't think that we need to print warning if ip is local at all. But if we want to do it we probably should do the following:
    - We shouldn't print warning in case of local Etherem network. It will add confusion to the users who try to play with our platform for the first time... So we should get id of Ethereum network and print warnings only for id=1 and 42.
    - We should print warning in self.out_f (not simply using print)
3. In the mpe_service_metadata.py we have the following definitions:
      - Two helper function ( ```is_valid_endpoint``` and ```is_private_endpoint```)
      - definition of ```MPEServiceMetadata```
      - Two helper functions (```load_mpe_service_metadata```, ```mpe_service_metadata_from_json```)

I would propose to move all helper function to the end. Or maybe put ```is_valid_endpoint``` and ```is_private_endpoint``` into utils.py or utils_endpoint.py.

ferrouswheel:
Yes it makes sense to silence the warning for local Ethereum networks and will use out_f (MPEServiceMetadata doesn't have a base class, can you advise which class I should inherit from to get `out_f`?).

Happy to move the utility methods out of the method too.

The reason I'm concerned about private hosts, is that it could be used as a way for someone to publish a service that would get clients to call resources in their internal network. If there is a grpc server somewhere in a LAN, and somebody writes an app that makes SingularityNET calls, they could later change the endpoint to a private address and then exfiltrate data to an external host ...(or if a person has their ethereum private key stolen by an attacker). But it's probably unlikely to be worth the effort and I am probably being paranoid. Besides, just validating it with snet-cli won't prevent people publishing private addresses.

The other reason is to avoid new developers accidentally publishing their local development IP to the registry instead of the correct one, and this could leak knowledge about private network configuration.

ferrouswheel:
Ok - I have removed the code related to checking private endpoints.

Now that I think about the security issue, it makes more sense for our SDK to protect against calling private IPs, and to allow an SDK user to override that if they want (maybe using a whitelist of allowed networks). cc @vforvalerio87

astroseger:
Now there is a conflict with snet_cli/utils_ipfs.py because of #180. Could you resolve them. And I will merge it.

ferrouswheel:
Conflicts resolved :rocket:

PR #180, opened 2019-01-21
Validate IPFS multihash

### LINKED ISSUE #91: check hash of files from IPFS
We should check hash of files from IPFS. Otherwise we can be attacked if ipfs client is compromised.
Function which read IPFS is here:
https://github.com/singnet/snet-cli/blob/49c6dbaca8b8aace1bdd77dbc1f6c0ae0397d0f6/snet_cli/utils_ipfs.py#L27-L30

ferrouswheel:
Hidden behind a conditional because it requires our gateway to allow queries on raw blocks.

We currently get a 400 error (bad request) when hitting the gateway, e.g. http://ipfs.singularitynet.io/api/v0/block/get?stream-channels=true&arg=QmZFBAAEx2bNA5aD3yBoaMviZ2Dn9qe8U7QGp21VjvB7ts

Assuming it works, this should close #91


ferrouswheel:
@astroseger Yup, good point! I hadn't completed this implementation as I hadn't been able to test against our public ipfs gateway.

However as mentioned in this issue on the daemon repo https://github.com/singnet/snet-daemon/issues/197, that implementation is also incorrect.

We need to get the content from the raw block, and then hash that block and extract the content from it.

Unfortunately it is not possible to robustly use cat and validate with the hash.

ferrouswheel:
Ok, this was a little more involved than anticipated. Python isn't great for ipfs support, so I had to implement my own raw block deserialisation.

This only works on the current merkledag and unixfs formats, so may stop working if all of https://github.com/ipld/specs/ is completed and made the default (The bit I've implemented is just the column on the right of the diagram and works on the latest ipfs version)

Before making a release with this, we need our snet gateway to accept "get block" api requests.

astroseger:
I would propose to exclude *_pb2 files and auto generated them during the installation.

ferrouswheel:
Okay, I wasn't sure if we had any other protoc file to compile, and since
these are unlikely to ever change I thought this would introduce the fewest
dependencies and avoid additional build and packaging steps.

If you're happy to have those things introduced, I can add them.

astroseger:
Actually we have two more .proto files which is better to "compile" in setup.py: ```control_serivce.proto``` and ```state_service.proto``` https://github.com/singnet/snet-cli/tree/master/snet_cli/resources/proto . Currently we are compiling them dynamically, which is fine but unnecessary. It would be better to move them in setup.py as well.

So I propose to compile your .proto files in setup.py. And I will make separe PR for statically compile  ```control_serivce.proto``` and ```state_service.proto```.

ferrouswheel:
Compilation of protobuf happens as a post install step in `setup.py` now. It also compiles `control_service.proto` and `state_service.proto` (but I haven't changed the dynamic compilation code that installs them to the user directory).

PR #182, opened 2019-01-22
Point the version to the latest contract 0.2.6

anandrgitnirman:


astroseger:
It is not enough. We need new NPM release of token contract (2.0.1 for example) with address of token contract in robsten network.

Issue #183, opened 2019-01-22
Encoding issues


arturgontijo:
If I run `snet` in a Docker container I get an encoding error:

```
# snet organization info ramonduraes
Error: 'ascii' codec can't encode character '\xe3' in position 32: ordinal not in range(128)
If you want to see full Traceback then run:
snet --print-traceback [parameters]

# locale
LANG=
LANGUAGE=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
```

I can set `locale` in Dockerfile and also force it within python:

```
# PYTHONIOENCODING=utf-8 snet organization info ramonduraes

Organization Name:
 - Ramon Durães

Organization Id:
 - ramonduraes

Owner:
 - 0xCEb196e0236C5B4EE62d5C87692284FBd52fCBD0
```

I'd like to suggest an `UnicodeEncodeError` handler.

One approach is using `stdout.buffer`, with it `snet` will become console's encoding independent.
Example [commands.py#L39](https://github.com/singnet/snet-cli/blob/master/snet_cli/commands.py#L39):
```
    def _printout(self, message):
        if self.out_f is not None:
                try:
                        print(message, file=self.out_f)
                except UnicodeEncodeError:
                        sys.stdout.buffer.write(message.encode("utf-8"))
```

astroseger:
was fixed by #184

ferrouswheel:
I am concerned we are introducing a problem for people. If the console doesn't support UTF-8, we shouldn't force it. I think the correct option is to configure the terminal correctly.

One way to avoid error would be to use `message.encode('ascii', 'ignore')`, while not forcing utf-8 encoding on a terminal that doesn't support it. This throws away characters that can't be represented.

A better way to avoid error is to use [unidecode](https://pypi.org/project/Unidecode/) to replace undisplayable unicode characters with similar ascii characters.

arturgontijo:
> I am concerned we are introducing a problem for people. If the console doesn't support UTF-8, we shouldn't force it. I think the correct option is to configure the terminal correctly.

[...]

PR #184, opened 2019-01-22
[fix] UnicodeEncodeError handler.

arturgontijo:
Related #183.

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -38,7 +38,7 @@ class Command(object):
     def _printout(self, message):
         if self.out_f is not None:
             try:
-                message += "\n"
+                message = str(message) + "\n"
                 self.out_f.write(message)
             except UnicodeEncodeError as e:
                 if isinstance(self.out_f, type(sys.stdout)):
@@ -49,7 +49,7 @@ class Command(object):
     def _printerr(self, message):
         if self.err_f is not None:
             try:
-                message += "\n"
+                message = str(message) + "\n"
                 self.err_f.write(message)
             except UnicodeEncodeError as e:
                 if isinstance(self.err_f, type(sys.stderr)):
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -20,7 +20,6 @@ import secrets
 import string
 
 
-
 class Command(object):
     def __init__(self, config, args, out_f=sys.stdout, err_f=sys.stderr):
         self.config = config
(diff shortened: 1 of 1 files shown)


astroseger:
It seems there is a small problem which I haven't seen before... In ```_printout``` we should write in self._out_f, not in stdout...

astroseger:
(it seems there is again a bug with circleCI, you should remove snet-cli from your circleCI... )

arturgontijo:
> It seems there is a small problem which I haven't seen before... In `_printout` we should write in self._out_f, not in stdout...

The direct way is to use `self.out_f.buffer.write((message + "\n").encode("utf-8"))`.
But it will break if user sends `self.out_f` as something else then `sys.stdout`(`io.TextIOWrapper`).

I'm investigating a better way to fix this.

> (it seems there is again a bug with circleCI, you should remove snet-cli from your circleCI... )

Thanks, will remove my fork's webhook.

astroseger:
1. There is code duplication in _printout and _printerr. I propose to make a new function _print(f) and put duplucated code here
2. I prose to use "raise" (which will raise exception again) instead of "_error"
3. (very optional) We could move ```message = str(message) + "\n"``` outside the try region.

snet_cli/commands.py
```diff
@@ -38,11 +37,25 @@ def _ensure(self, condition, message):
 
     def _printout(self, message):
         if self.out_f is not None:
-            print(message, file=self.out_f)
+            try:
+                message = str(message) + "\n"
+                self.out_f.write(message)
+            except UnicodeEncodeError as e:
+                if isinstance(self.out_f, type(sys.stdout)):
```
astroseger: Why do you need this if?

astroseger: so you should use hasattr function

arturgontijo: Ok.
Is there a reason to use `hasattr(self.out_f, "buffer")` or can I use `getattr(self.out_f, "buffer", None)`?

astroseger: ```hasattr``` is a little bit cleanear because if attribute exists but it is ```None``` then ```getattr``` will return None, but hasttr will return true... But in this case of course result will be the same...

arturgontijo: I'm asking because `hasattr` will call `getattr` to return `True or False`.
But maybe using `hasattr` will make this part concise with other code parts.
I'm ok to use it BTW.
(I'm working with frontend now, so my head is far away from python. =/)

PR #190, opened 2019-01-26
Keystore

### LINKED ISSUE #187: support keystore files as an identity.
We should have "keystore file" as one of possible identity types.

tiero:
New arguments for `snet identity create` and new `IDENTITY_TYPE` has been added

 Now you can use an encrypted keystore JSON file (the same MyEtherWallet provides so to speak) as the source for your private key. You will be prompted to unlock with your password the keystore before signing. 

This closes #187


tiero:
Please, review @astroseger @vforvalerio87

tiero:
Sure I thought extensively about I, my rationale behind was that since it's already ok for us storing both mnemonic or private key plain text, I thought that having this proactive step would have broken the flow, especially for scripts as you pointed out. 
I would do the same way we do with ledger and trezor, also them need interactive step for each sign.

About II, I basically re-implemented only the decode function from https://github.com/ethereum/pyethereum, in order to not have an additional external library.

tiero:
Pardon, About || you are right cause actually we already have web3.py, so I will remove custom util code and just use that

astroseger:
Probably for ledger and trezor we should add the same trick: add possibility to actually save the password in .config (of course only as an option and with warnings)

tiero:
both Ledger and Trezor have no way to supply a password programmatically to unlock (for security reasons) so an human need to manually input it using the hardware keryboard.

At this point, I would manage the key-store as if it is a hardware wallet, an unlock via a user prompt input only. So I would drop `--keystore-password` on init

I think since we already provide multiple ways (privKey and mnemonic) as a way to have all "unlocked" , better for the keystore to be unlocked only when needed.

astroseger:
But why you don't like my proposition to make storing the password an option? 
If user is confident in his setup he can store his password (and use snet-cli in scripts).

astroseger:
I mean. I agree that from security reason it is better to not store the password (and we will push users to not do it). But it might be useful in some setups. 
And you already have it, so it is quite straightforward to simply make storing the password as an option...

astroseger:
The problem is following: There is one command which service provider will probably need to run from scripts: ```snet treasurer claim-expired``` (see step 10 in tutorial: https://github.com/singnet/dev-portal/blob/master/tutorials/publish/index.md).  So I think it is better to leave possibility to store password (of course, only as an option).

But probably your are right in the sense that we can use "key-identity" in this case. If service provider is confident in his setup he can extract private key from keystore file and make key-identity...

tiero:
Exactly that, if he needs a particular setup where having the private key in plain text is crucial, just use key or mnemonic directly, without adding more complexity.

I would handle keystore in the same way as hardware wallets. I know I already have it, but better to decrease the surface of attack and foster good practices to avoid user making mistakes.

tiero:
With this change, I removed custom function to decode in favor of web3.py built-in util.

Also, I no longer allow users to store the password in plain text, handling the KeyStore in the same manner as Ledger and Trezor

astroseger:
1. I propose to ask for password only once (per snet run) and store it in self.private_key  (and if self.private_key exists then reuse it). Sometimes we have several transactions per session, for example in ```snet treasurer claim-all``` (we first call sign_message_after_soliditySha3 in order to sign message to the daemon, and then we sign, possibly, multiply transactions).  It shouldn't be a security problem, because if memory is compromised we are in troubles in any way.  It will also remove copy-paste in ```KeyStoreIdentityProvider.transact``` and ```KeyStoreIdentityProvider.sign_message_after_soliditySha3```

2.  error message in ```KeyStoreIdentityProvider.__init__``` is outdated

3. In all identity providers which use private key  (KeyIdentityProvider, MnemonicIdentityProvider, KeyStoreIdentityProvider) we have code duplication. I propose to make two functions for sake of perfectionism (and reuse them everywhere): 
```
 def transact_with_private_key(w3, private_key, transaction, out_f):
        raw_transaction = w3.eth.account.signTransaction(
            transaction, private_key).rawTransaction
        return send_and_wait_for_transaction(raw_transaction, w3, out_f)

def sign_message_with_private_key(w3, message):
       h = defunct_hash_message(message)
        return w3.eth.account.signHash(h, private_key).signature
```

tiero:
agreed for 2 and 3, for 1 are you saying that whatever is the first function being called (transact or sign_message...) we save the private key, so we should check if not None, so we do not decrypt the keystore again in current session?

astroseger:
> agreed for 2 and 3, for 1 are you saying that whatever is the first function being called (transact or sign_message...) we save the private key, so we should check if not None, so we do not decrypt the keystore again in current session?

exactly!

tiero:
@astroseger Fixed as agreed, the only different thing is about your `transact_with_private_key` example, where I left the broadcast of transaction outside the helper I called `sign_transaction_with_private_key` so it remains functional, without side effects (blockchain)

astroseger:
Should I merge?

PR #200, opened 2019-01-31
Changes to allow using snet-cli as a library to simulate grpc call

astroseger:
This PR is only for discussion.

Changes which allow using snet-cli as a pseudo sdk in: https://github.com/astroseger/simple_snet_sdk

* add ```MPEClientCommand.call_server_statelessly_with_params```
* rewrite ```utils_proto.py``` in order to isolate ```import_protobuf_from_dir_get_all``` which we need to simulate GRPC in pseudo sdk.

It should be noted that the last change (modification of ```utils_proto.py```) is optional because  it can be done in ```https://github.com/astroseger/simple_snet_sdk```

### DIFF
--- a/snet_cli/mpe_client_command.py
+++ b/snet_cli/mpe_client_command.py
@@ -63,7 +63,7 @@ class MPEClientCommand(MPEChannelCommand):
 
         return params
 
-    
+
     def _transform_call_params(self, params):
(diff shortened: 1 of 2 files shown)


astroseger:
was partly moved to #210

Issue #204, opened 2019-02-03
snet-cli assumes all proto types are defined in the same module


ferrouswheel:
Damn, #84 is still not completely resolved.

We have a big assumption about how we dynamically load python grpc modules.

```
$ snet --print-traceback client call 912 0.000001 34.216.72.29:6201 FindFace '{"file@ImageRGB": "tests/test_images/laos.jpg"}'
unspent_amount_in_cogs before call (None means that we cannot get it now):100
Traceback (most recent call last):
  File "/home/joel/.virtualenvs/faces/bin/snet", line 11, in <module>
    load_entry_point('snet-cli', 'console_scripts', 'snet')()
  File "/home/joel/work/snet-cli/snet_cli/__init__.py", line 20, in main
    getattr(args.cmd(conf, args), args.fn)()
  File "/home/joel/work/snet-cli/snet_cli/mpe_client_command.py", line 217, in call_server_statelessly
    response = self._call_server_via_grpc_channel(grpc_channel, current_nonce, current_amount + self.args.price, params, service_metadata)
  File "/home/joel/work/snet-cli/snet_cli/mpe_client_command.py", line 96, in _call_server_via_grpc_channel
    stub_class, request_class, response_class = self._import_protobuf_for_channel()
  File "/home/joel/work/snet-cli/snet_cli/mpe_client_command.py", line 93, in _import_protobuf_for_channel
    return import_protobuf_from_dir(channel_dir, self.args.method, self.args.service)
  File "/home/joel/work/snet-cli/snet_cli/utils_proto.py", line 22, in import_protobuf_from_dir
    is_found, rez = _import_protobuf_from_file(grpc_pyfile, method_name, service_name);
  File "/home/joel/work/snet-cli/snet_cli/utils_proto.py", line 60, in _import_protobuf_from_file
    request_class      = getattr(pb2, method.input_type.name)
AttributeError: module 'face_detect_pb2' has no attribute 'ImageRGB'
```

`ImageRGB` is a message defined in `face_common_pb2`, which is imported by `face_detect_pb2_grpc`.

ferrouswheel:
Actually I think I have a fix that just uses the `._concrete_class` instead of `getattr(..)`

Now I just need to remove grpc streams and the face-services will be close to operational again :-)

PR #205, opened 2019-02-04
Misc proto/grpc fixes

ferrouswheel:
- allow modifiers to appear on keys in nested messages
- use correct class for request and response, #204


ferrouswheel:
I'm currently trying to just get any of my services operational for the beta so want to merge this right away. Happy to help with some testing and clean up after. #206 #207

Issue #208, opened 2019-02-07
Support domain name based services from snet-cli


raamb:
The client only allows service calls to ip:port. This is a problem if a daemon uses a domain name. One scenario is if a daemon is getting its SSL certificates from LetsEncrypt using a domain name.

ferrouswheel:
As far as I can tell there is nothing that explicitly prevents using a domain name. What version are you using and what is the error you are getting?

raamb:
Current snet-client does not do a name lookup per Sergey.

ferrouswheel:
Why does snet-client need to do a name lookup?

Opening a grpc connection using a hostname should take care of that.

astroseger:
Yes. It seems you are right, grpc should make hostname lookup by itself (https://grpc.io/grpc/core/md_doc_naming.html). Which raises some question about security... We really should start to use secure channels asap.

astroseger:
was fixed by #210 (in general this issue was about support https...)

Issue #201, opened 2019-02-01
Call service with 'org_id + service_id + method_name' not whit 'channel_id + method'


astroseger:
In the current version ```snet client call``` has the following parameters: ```channel_id, price, endpoint, method, params```.
It might be more convenient for users to have the following parameters: ```price, org_id, service_id, method```.  
* we could remove ```price```. But we probably should keep it for cross check.
* ```endpoint``` could be made optional. By default, we read endpoint from metadata and use round-robin in case of multiply endpoints
* ```channel_id``` is optional. We need it only in case we have several initialized channel for the same service. 

To implement it, we need to have a way to find all initialized channels which was open for specific service. We already have it  (see ```MPEChannelCommand.print_initialized_channels_filter_group```). However it might be not the fastest way (in case of many initilized channels). But I guess it is fast enough for now...

ferrouswheel:
Please make it possible to do `snet client call ORG SERVICE_NAME METHOD_NAME`.

Don't require price, if no price is specified you can ask the user to confirm, but it's really frustrating to have to look it up.

astroseger:
Should user specify the price? It is a question.
I prefer that user specify the price, because otherwise he can be attacked by malicious service provider.

ferrouswheel:
The CLI can always ask for the user to confirm they are okay with the current price.

Currently I have to look up the price in the `service_metadata.json` and it's frustrating to remember (especially because I have to covert between cogs and AGI).

arturgontijo:
Remove price would be nice (with prompt price value question, like"The service's price is 10 COGs, proceed?") and keep the `--yes` to avoid this if wanted.

astroseger:
was closed by #210

PR #210, opened 2019-02-14
Hide channel id and endpoint from user and support https

astroseger:
This PR solves #201 and #208. It adds proper support for secure channels (https) and it significantly simplifies the life for users by hiding channel_id (and endpoints).

Example of workflow:
```
snet network ropsten

# open new channel only if you haven't done it already 
# It will not open new channel if you already have one!
snet channel open-init DappTesOrganization DappTesthttpsService 0.00001 +8days

# add funds if needed (for service which was opened for specific service)
snet channel extend-add-for-service DappTesOrganization DappTesthttpsService --amount 0.00001 --expiration +8days

# print list of initialized channels 
snet channel print-initialized

# make a call (you can use -y to remove price warning)
snet client call DappTesOrganization DappTesthttpsService mul '{"a":12,"b":7}'

# try claim timeout for ALL your channels (channels which have your as a sender)
snet channel claim-timeout-all
```

This PR does the following:
- change directories struct in ```~/.snet/mpe_client``` in order to make it more similar to @vforvalerio87 SDK.  This new directories structure allows hiding channel_id from user.
     - Store compiled protobuf for service with ```<org_id>``` and ```<service_id>``` in ```~/.snet/mpe_client/<mpe_address>_<registry_address>/<org_id>/<service_id>/service```
    - Store initialized channels for this service in : ```~/.snet/mpe_client/<mpe_address>_<registry_address>/<org_id>/<service_id>/channels```
- By default ```snet channel open-init``` will not open new channel if old channel for this service already exists. It will simply initialize existed channel.
- ```snet client call``` has the following positional parameters: ```<org_id> <service_id> <method> <params>```. User will need to manually specify channel_id only in case of multiply channels initialized for the same service (which, by default, is prevented  in ```snet channel open-init```).
- make ```snet client call``` more faster by removing unnecessary blockchain call.
- Remove price from list of obligatory parameter in ```snet client call```. Instead, as was  proposed by @arturgontijo, we have interactive confirmation. We have possibility to skip this confirmation by ```--yes/-y```.
- Change output of ```snet channel print-initialized``` to include org_id and service_id and remove unnecessary group_id
- rename ```snet channel print-initialized-filter-group``` to ```snet channel print-initialized-filter-service``` because now it will print all initialized channels for whole service (not only for the specifc group). Change output to more informative similarly to ```snet channel print-initialized```.
- Add ```snet channel claim-timeout-all``` function which check ALL channels for the given identity and claim timeout for all expired channels.
- Add ```snet channel extend-add-for-service``` function which add funds and extend expiration data for the channel which was opened for specific channel
- (small change) In ```snet service update-metadata``` check that group_id or payment_address haven't been changed. We have possibility to force this change by ```--force```
- Add tests for all new functions and for some old ones (```snet client call-lowlevel```, ```snet client get-channel-state```)

### DIFF
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -569,6 +569,7 @@ def add_mpe_client_options(parser):
     add_p_org_id_service_id(p)
     add_p_set1_for_call(p)
     p.add_argument("--channel-id", type=int, help="channel_id (only in case of multiply initilized channels for the same payment group)")
+    p.add_argument("--yes", "-y", action="store_true", help="skip interactive confirmation of call price", default=False)
 
 
     p = subparsers.add_parser("call-lowlevel", help="Low level function for calling the server. Service should be already initilized.")
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -487,7 +487,7 @@ def add_mpe_channel_options(parser):
     add_p_channel_id(p)
     add_p_set_for_extend_add(p)
 
-    p = subparsers.add_parser("extend-add-for-service", help="Set new expiration for the channel which was initilized for the given service")
+    p = subparsers.add_parser("extend-add-for-service", help="Set new expiration and add funds for the channel which was initilized for the given service")
     p.set_defaults(fn="channel_extend_and_add_funds_for_service")
     add_p_service_in_registry(p)
     add_p_set_for_extend_add(p)
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -344,17 +344,22 @@ class AppendPositionalAction(argparse.Action):
 def add_p_mpe_address_opt(p):
     p.add_argument("--multipartyescrow-at", "--mpe", default=None,  help="address of MultiPartyEscrow contract, if not specified we read address from \"networks\"")
 
+def add_p_registry_address_opt(p):
+    p.add_argument("--registry-at", "--registry", default=None, help="address of Registry contract, if not specified we read address from \"networks\"")
+
 
 def add_p_metadata_file_opt(p):
     p.add_argument("--metadata-file", default="service_metadata.json", help="Service metadata json file (default service_metadata.json)")
+
 
 def add_p_org_id_service_id(p):
     add_p_org_id(p)
     p.add_argument("service_id",      help="Id of service")
 
+
 def add_p_service_in_registry(p):
     add_p_org_id_service_id(p)
-    p.add_argument("--registry-at", "--registry", default=None, help="address of Registry contract, if not specified we read address from \"networks\"")
+    add_p_registry_address_opt(p)
 
 
 def add_mpe_account_options(parser):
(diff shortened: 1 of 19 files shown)


arturgontijo:
Wow...this one was awesome, @astroseger!

snet_cli/arguments.py
```diff
@@ -550,26 +565,34 @@ def add_mpe_client_options(parser):
     subparsers = parser.add_subparsers(title="Commands", metavar="COMMAND")
     subparsers.required = True
 
-    def add_p_deal_with_response(p):
+    def add_p_set1_for_call(p):
```
raamb: Can we make the name more descriptive. Not sure what set1 means in this context.

astroseger: It is intentional. This function represent common parameters for ```call``` and ```call-lowlevel```. But these parameters too diverse to find a common name for them, and I don't want anybody to reuse this function, so I called it "set1"...

snet_cli/mpe_channel_command.py
```diff
@@ -12,91 +13,111 @@
...
+            metadata.save_pretty(os.path.join(service_dir, "service_metadata.json"))
         except:
             # it is secure to remove channel_dir, because we've created it
-            shutil.rmtree(channel_dir)
+            shutil.rmtree(service_dir)
             raise
 
+    def _init_or_update_service_from_metadata(self, metadata):
+        tmp_dir = tempfile.mkdtemp()
+        shutil.rmtree(tmp_dir)
```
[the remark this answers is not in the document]
astroseger: yes... It is bad design. I will rewrite this function using tempfile.TemporaryDirectory pattern...

snet_cli/mpe_client_command.py
```diff
@@ -92,22 +90,21 @@ def _transform_call_params(self, params):
...
+    def _call_server_via_grpc_channel(self, grpc_channel, channel_id, nonce, amount, params, service_metadata):
+        stub_class, request_class, response_class = self._import_protobuf_for_service()
 
         request  = request_class(**params)
         stub     = stub_class(grpc_channel)
         call_fn  = getattr(stub, self.args.method)
 
         if service_metadata["encoding"] == "json":
-            switch_to_json_payload_econding(call_fn, response_class)
+            switch_to_json_payload_encoding(call_fn, response_class)
```
raamb: Can we not remove this? We dont use JSON encoding any more right

snet_cli/mpe_service_command.py
```diff
@@ -108,6 +108,19 @@ def publish_service_with_metadata(self):
...
+        # first we check that we do not change payment_address or group_id in existed payment groups
+        if (not self.args.force):
+            old_metadata = self._get_service_metadata_from_registry()
+            new_metadata = load_mpe_service_metadata(self.args.metadata_file)
+            for old_group in old_metadata["groups"]:
+                if (new_metadata.is_group_name_exists(old_group["group_name"])):
+
+                    new_group = new_metadata.get_group(old_group["group_name"])
+                    if (new_group["group_id"] != old_group["group_id"] or new_group["payment_address"] != old_group["payment_address"]):
+                        raise Exception("You are trying to change group_id or payment_address in group '%s'.\n"%old_group["group_name"] +
```
raamb: Would it make sense to auto claim the channels before we complete this operation? We dont need to do this for the beta we can do it later. For the beta we could add a line asking them to claim the funds before the change.

snet_cli/mpe_client_command.py
astroseger: But I'm not removing it.. I fix typo in "encoding" name.

snet_cli/mpe_service_command.py
astroseger: But we cannot auto claim channels for users... This line prevents service providers to change group_id or payment_address by mistake.

astroseger:
@raamb I've improved work with temporary directories. Could you approve :) ?

Issue #29, opened 2018-08-11
Code generation


ldub:
CLI should have the ability to generate client stubs for calling SingularityNET AI Services in a variety of supported programming languages

Depends on #28

astroseger:
It should be noted that it will partly solve issue with snet-cli being relatively slow #141 (if we properly generate stubs)

PR #219, opened 2019-02-24
Simple call stub generator for calling 'snet client call' from a python code

astroseger:
This PR partly deals with https://github.com/singnet/snet-cli/issues/29 (for python).
It adds simple call stub generator for calling services from python, using snet-cli infrastructure.

For example you can write the following code for calling ```snet/example-service```

```python
from snet_cli.call_stub_generator import call_stub_generator

stub = call_stub_generator("snet", "example-service", "add")

# parameters should be a dictionary (not a json!), so it can have binary fields.
params = {"a":1, "b":2}

rez = stub.call_server_statelessly_with_params(params)
print(rez)
``` 
- Parameters are passed in form of a dictionary. It is relative easy to pass parameters directly in Request format, but I don't think it is necessary .  
- It should be noted that this PR is independent from  #218 but after #218 you will be able to run this python code without any channel initialization in snet-cli (but of course only if you already have a funded channel).

### DIFF
--- a/snet_cli/call_stub_generator.py
+++ b/snet_cli/call_stub_generator.py
@@ -5,4 +5,9 @@ from snet_cli.utils import DefaultAttributeObject
 def call_stub_generator(org_id, service_id, method, **kwargs):
     conf = Config()
     args = DefaultAttributeObject(org_id = org_id, service_id = service_id, method = method, yes = True, **kwargs)
-    return MPEClientCommand(conf, args)
+    class Stub:
+        def __call__(self, params):
+            return self.client.call_server_statelessly_with_params(params)
+    stub = Stub()
+    stub.client = MPEClientCommand(conf, args)
+    return stub
--- a/snet_cli/call_stub_generator.py
+++ b/snet_cli/call_stub_generator.py
@@ -1,0 +1,8 @@
+from snet_cli.mpe_client_command import MPEClientCommand
+from snet_cli.config import Config
+from snet_cli.utils import DefaultAttributeObject
+
+def call_stub_generator(org_id, service_id, method, **kwargs):
+    conf = Config()
+    args = DefaultAttributeObject(org_id = org_id, service_id = service_id, method = method, yes = True, **kwargs)
+    return MPEClientCommand(conf, args)
--- a/test/functional_tests/script14_stub_tests.py
+++ b/test/functional_tests/script14_stub_tests.py
@@ -1,0 +1,6 @@
+from snet_cli.call_stub_generator import call_stub_generator
+
+stub = call_stub_generator("testo", "tests", "classify")
+
+rez = stub.call_server_statelessly_with_params({})
+print(rez)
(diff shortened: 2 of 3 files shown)


ferrouswheel:
Is there any reason we can't rename `stub.call_server_statelessly_with_params` to `stub.call`? Or even better, implement `__call__` on the object returned by `call_stub_generator`. Then you could just do:

```
add_stub = call_stub_generator("snet", "example-service", "add")
rez = add_stub({"a":1, "b":2})
```

Python community tends to prefer concise APIs over detailed method names (the details should go in the docstring for the method).

astroseger:
@ferrouswheel 
Done! It is implemented via ```__call__```.  So example code will be as following:

```python
from snet_cli.call_stub_generator import call_stub_generator

stub = call_stub_generator("snet", "example-service", "add")

# parameters should be a dictionary (not a json!), so it can have binary fields.
params = {"a":1, "b":2}

rez = stub(params)
print(rez)
```

astroseger:
@raamb @vforvalerio87  what do your think? should I merge it?

astroseger:
@raamb 
I think that there is a value in both ```SDK``` and ```call-stub-generator```. 
stub generator is simpler and it is actually a part of snet-cli (which have obvious prons and cons), and it is more suitable for applications which dynamically call many services. ```SDK``` might be more useful for someone who write the client for his own service and want to have more grpc like experience. 

Call example service with stub generator
```
from snet_cli.call_stub_generator import call_stub_generator

stub = call_stub_generator("snet", "example-service", "add")
params = {"a":1, "b":2}
rez = stub(params)
print(rez)
```

Call example service with SDK. It is fully ```grpc``` like expirience, but you need to provide the name of protobuf file and the name of Request class, so it is not suitable for many unknown service. And of course you cannot rely on ```snet-cli``` as an identity provider.
```
from snet_sdk import Snet

snet = Snet(private_key="...")
calculator = snet.client("snet", "example-service")
calc_stub = calculator.grpc.example_service_pb2_grpc.CalculatorStub(calculator.grpc_channel)
calc_request = calculator.grpc.example_service_pb2.Numbers(a=20, b=30)
print(str(calc_stub.add(calc_request).value))
```

astroseger:
@raamb should I merge it ;) ?

raamb:
Apologies for the delay in my response @astroseger. I would still like to spend some more time on understanding why this cant be made a part of the SDK. Lets discuss that offline and get back to this.

astroseger:
@raamb @vforvalerio87 
It should be noted that the most significant different between SDK and stub generator is who is keeping the Ethereum identity. 
In this simple stub-generator we relay on snet-cli to keep our identity. In SDK we should provide our identity  to SDK... 
It means that we never be able to simulate stub-generator experience with current SDK.

astroseger:
@vforvalerio87 @raamb Do we need this stub generator? It might be the lightweight alternative for the SDK. But I'm not sure.

Issue #108, opened 2018-11-12
Autocompletion with TAB


astroseger:
I would propose to support standard autocompletion pattern with tab.

For example:
* complete "snet co" after pressing tab to "snet contract"
* complete "snet contract M" after pressing tab to "snet contract MultiPartyEscrow"
* etc...

It should be a way to do it very easily since we are using standard argparse library. However I coudn't find a fast way to do it....

arturgontijo:
I'd like to suggest `argcomplete` as we are using `argparse`.
The implementation is simple as:
```
import sys
import argcomplete

from snet_cli import arguments
from snet_cli.config import conf

__version__ = "0.1.6"


def main():
    argv = sys.argv[1:]
    parser = arguments.get_root_parser(conf)
    argcomplete.autocomplete(parser)
    try:
        args = parser.parse_args(argv)
...
```
The downside is that user must enable the global completion by:
`register-python-argcomplete snet >> ~/.bashrc`
or
`echo 'eval "$(register-python-argcomplete snet)"' >> ~/.bashrc`

And then:
`source ~/.bashrc`

IMO put this at README will be better than inside `snet` code.

Dinesh2521:
As @arturgontijo mentioned, argcomplete should be  fine for implementing auto completion. It will help with commands/sub commands auto completion.

There are 2 types of auto completion, which we may require in our CLI
i. Commands/Sub commands - comes by default with argcomplete package. 
ii. Choices/Options we can pass as arguments (Few things comes by default, but we may need to write **custom completers for our CLI specific choices/options**.)

Here is the package for more details
https://pypi.org/project/argcomplete/

**good**
- works well with Linux. Yet to validate on MacOS(Mostly it may work on MacOS as well)
- commands auto complete can be enabled easily 
- it can be used for main commands and sub commands as well
- we can do both arguments and option completers
- it will give us options to write custom completers, so that we can pull choices list from the services, local files, or any other sources 


**bad**
- will not work with Windows OS. Need to modify the the source to enable support for windows
- doesn't offer all parameters auto complete, but we can extend it by using custom completers
- will not offer recommended options or suggestions like AWSCli or few other CLI programs. it does very basic.

raamb:
PR #222 provides this functionality now

PR #222, opened 2019-02-26
Autocompletion of snet commands with TAB (fix for #108)

Dinesh2521:
This change enables auto completion of snet commands with pressing of TAB key after we enter 3 char. We are using the argcomplete package, since our commands leverage argparse.

one key observation with argcomplete is, sometimes the auto completion registration scripts may not come with package installation. To avoid the issue, we can install that package as a prerequisite (steps are added to readme)

@raamb @astroseger can you please review the changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ You should have python with version >= 3.6.5 and pip installed.
 Additionally you should install the following packages:
 
 * libudev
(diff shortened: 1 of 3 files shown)


raamb:
@astroseger request your review for this

PR #224, opened 2019-02-28
Gas price strategies

### LINKED ISSUE #223: Use time based gas price strategies from web3.py
We should allow using different gas price strategies from web3.py: https://web3py.readthedocs.io/en/stable/gas_price.html

I think time based gas price strategies is the most convenient one.

astroseger:
fix #223 

This PR add possibility to use time based gas price strategies (https://web3py.readthedocs.io/en/stable/gas_price.html).
We can set gas price as following:
- gas price in Wei (integer)
- 'fast' - Transaction mined within ~60 seconds.
- 'medium' - Transaction mined within ~5 minutes.
- 'slow'  - Transaction mined within 1 hour.

I've set ```--gas-price=medium``` by default for all networks including mainnet. 
@raamb Maybe we should default it to ```fast``` for mainnet to improve user experience?

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -15,10 +15,9 @@ from snet_cli.utils import DefaultAttributeObject, get_web3, serializable, type_
 from snet_cli.utils_config import get_contract_address, get_field_from_args_or_session
 from snet_cli.identity import RpcIdentityProvider, MnemonicIdentityProvider, TrezorIdentityProvider, \
     LedgerIdentityProvider, KeyIdentityProvider, KeyStoreIdentityProvider
-import web3
+from web3.eth import is_checksum_address
 import secrets
 import string
-from web3 import middleware
 from web3.gas_strategies.time_based import fast_gas_price_strategy, medium_gas_price_strategy, slow_gas_price_strategy
 
 
@@ -348,7 +347,7 @@ class OrganizationCommand(BlockchainCommand):
             return []
         members = [m.replace("[", "").replace("]", "") for m in self.args.members.split(',')]
         for m in members:
-            if not web3.eth.is_checksum_address(m):
+            if not is_checksum_address(m):
                 raise Exception("Member account %s is not a valid Ethereum checksum address"%m)
         return members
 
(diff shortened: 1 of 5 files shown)


astroseger:
Sorry guys. Apparently there is a bug in middleware. Thanks God we have CI. 
Please wait with your review...

astroseger:
@arturgontijo @raamb 
I bypassed the bug with middleware (by not using it :) ). Please review it.

snet_cli/commands.py
```diff
@@ -18,6 +18,9 @@
 import web3
```
arturgontijo: `from web3.eth import is_checksum_address` instead?

snet_cli/commands.py
```diff
@@ -18,6 +18,9 @@
...
 import secrets
 import string
+from web3 import middleware
```
arturgontijo: Unused.

snet_cli/commands.py
```diff
@@ -83,11 +86,36 @@ def show(self):
         self._pprint({"version": get_cli_version()})
 
 
+class cached_gas_price_strategy:
```
arturgontijo: Suggestion: CamelCase.

snet_cli/commands.py
```diff
@@ -83,11 +86,36 @@ def show(self):
...
+    def __init__(self, gas_price_param):
+        self.gas_price_param = gas_price_param
+        self.cached_gas_price = None
+    def __call__(self, web3, transaction_params):
```
arturgontijo: `web3 -> w3` as it's in the old codes.

snet_cli/commands.py
```diff
@@ -96,8 +124,16 @@ def get_eth_endpoint(self):
...
 
-    def get_gas_price(self):
-        return int(get_field_from_args_or_session(self.config, self.args, "gas_price"))
+    def get_gas_price_param(self):
+         return get_field_from_args_or_session(self.config, self.args, "gas_price")
+
+    def get_gas_price_verbose(self):
+        # gas price is not given explicitly in Wei
+        if (self.w3.eth.gasPriceStrategy.is_going_to_calculate()):
+            self._printerr("# Calculating gas price. It might take ~60 seconds.")
```
arturgontijo: I test this like 5x...and the delay to find the gas price is about ~35s (Ropsten).
But even with this delay, the transaction seems to take the same time as before on Ropsten. =/
Is this the expected behavior?

astroseger: Yes. On ropsten you can set gas_price = 1 Wei, and it still will pass with the same speed.

This PR will affect only the mainnet. But it should simplify the life of users significantly, otherwise users will need to manually set the gas-price, but to do so user should understand what the gas price is...

PR #225, opened 2019-02-28
Metadata remove endpoints

astroseger:
This PR does following

- Add 'snet service metadata-remove-all-endpoints' command which remove all endpoints from metadata file
- remove '--force' option from 'snet service update-metadata' which fully remove possibility to accidentally change ```group_id``` or ```payment_address``` for existed services (in case of emergency it is always possible to remove service and add it again).

In case user try to update ```group_id``` or ```payment_address``` for existed service (by running ```snet service metadata-update```)  he will see the following error message:
```
You are trying to change group_id or payment_address in group '.....'.
It would make all open channels invalid.
You shoudn't use 'metadata-init' for existed service, because it reinitialize group_id
Please use 'metadata-set-model' for change your protobuf file
Please use 'metadata-remove-all-endpoints/metadata-add-endpoints to update endpoints
```

### DIFF
--- a/snet_cli/mpe_service_command.py
+++ b/snet_cli/mpe_service_command.py
@@ -60,6 +60,12 @@ class MPEServiceCommand(BlockchainCommand):
             metadata.add_endpoint(group_name, endpoint)
         metadata.save_pretty(self.args.metadata_file)
 
+    def metadata_remove_all_endpoints(self):
+        """ Metadata: remove all endpoints from all groups """
+        metadata = load_mpe_service_metadata(self.args.metadata_file)
+        metadata.remove_all_endpoints()
+        metadata.save_pretty(self.args.metadata_file)
+
     def metadata_add_description(self):
         """ Metadata: add description """
         service_description = {}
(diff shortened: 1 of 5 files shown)


snet_cli/mpe_service_command.py
```diff
@@ -109,18 +115,17 @@ def publish_service_with_metadata(self):
...
-                    new_group = new_metadata.get_group(old_group["group_name"])
-                    if (new_group["group_id"] != old_group["group_id"] or new_group["payment_address"] != old_group["payment_address"]):
-                        raise Exception("You are trying to change group_id or payment_address in group '%s'.\n"%old_group["group_name"] +
-                                         "You will make all open channels invalid.\n" +
-                                         "Use --force if you really want it.")
-
+        old_metadata = self._get_service_metadata_from_registry()
+        new_metadata = load_mpe_service_metadata(self.args.metadata_file)
+        for old_group in old_metadata["groups"]:
+            if (new_metadata.is_group_name_exists(old_group["group_name"])):
```
raamb: Can it happen that the group names change completely across a publish. i.e. I have g1 and g2 in old_metadata and g3 and g4 in new_metadata

astroseger: Yes, it is possible. I only check that ```group_id``` or ```payment_address``` haven't been changed in the same groups. It will work in most cases because almost everybody have only one `default_group`.

We also probably shoudn't forbid to remove payment groups...

arturgontijo:
Another thing, if user keeps adding endpoint with:
`snet service metadata-add-endpoints https://test.com:6001`
`snet service metadata-add-endpoints https://test.com:6001`

astroseger:
@arturgontijo 
I've added function `snet service metadata-update-endpoints` which remove all endpoints from the given group and add new ones (group_name should be specified only in case of multiply payment_groups)

And I didn't understand you question about adding endpoints with `snet service metadata-add-endpoints`. If you try to add already exited endpoint snet-cli will fail with the error.

arturgontijo:
```
$ snet service print-metadata snet example-service > service_metadata.json 
$ cat service_metadata.json 
{
    "version": 1,
    "display_name": "SingularityNET Example Service",
    "encoding": "proto",
    "service_type": "grpc",
    "payment_expiration_threshold": 40320,
    "model_ipfs_hash": "QmWsc9QUD6emiAnDuuFFMU776CRTGYba5WD9XW3e2BVqVf",
    "mpe_address": "0x7E6366Fbe3bdfCE3C906667911FC5237Cc96BD08",
    "pricing": {
        "price_model": "fixed_price",
        "price_in_cogs": 1
    },
    "groups": [
        {
            "group_name": "default_group",
            "group_id": "0GwlhBaY9ke6gbX3e7CW6G/jKu9jmr5iGtKqpwZP4U4=",
            "payment_address": "0x501e8c58E6C16081c0AbCf80Ce2ABb6b3f91E717"
        }
    ],
    "endpoints": [
        {
            "group_name": "default_group",
            "endpoint": "https://bh.singularitynet.io:7052"
        }
    ],
    "service_description": {
        "description": "Simple arithmetic service compatible with SingularityNET.",
        "url": "https://github.com/singnet/example-service"
    }
}
$ snet service metadata-add-endpoints http://localhost.com
$ cat service_metadata.json 
{
    "version": 1,
    "display_name": "SingularityNET Example Service",
    "encoding": "proto",
    "service_type": "grpc",
    "payment_expiration_threshold": 40320,
    "model_ipfs_hash": "QmWsc9QUD6emiAnDuuFFMU776CRTGYba5WD9XW3e2BVqVf",
    "mpe_address": "0x7E6366Fbe3bdfCE3C906667911FC5237Cc96BD08",
    "pricing": {
        "price_model": "fixed_price",
        "price_in_cogs": 1
    },
    "groups": [
        {
            "group_name": "default_group",
            "group_id": "0GwlhBaY9ke6gbX3e7CW6G/jKu9jmr5iGtKqpwZP4U4=",
            "payment_address": "0x501e8c58E6C16081c0AbCf80Ce2ABb6b3f91E717"
        }
    ],
    "endpoints": [
        {
            "group_name": "default_group",
            "endpoint": "https://bh.singularitynet.io:7052"
        },
        {
            "group_name": "default_group",
            "endpoint": "http://localhost.com"
        }
    ],
    "service_description": {
        "description": "Simple arithmetic service compatible with SingularityNET.",
        "url": "https://github.com/singnet/example-service"
    }
}
```
Is it possible to have same `group_name` with more than one endpoint?
Is so, maybe a list here would be better (but this is minor).

astroseger:
Yes, maybe list might be better. But we've decided to keep it simpler (and more similar to simple DB). And in order to change it we should change all other applications (dApp and daemon). So I propose to not change format of metadata :) .

astroseger:
@raamb do you approve?

snet_cli/mpe_service_command.py
raamb: So shouldnt we do the same check where group names have been changed completely? In that scenario too open channels would become invalid.

astroseger: In principle we could forbid to removing payment_group (via snet-cli). There is always a possibility to simply remove service and register it again....
So should I forbid to removing group?

raamb: Preventing developers from removing payment groups seems restrictive for me. Imagine that I had set up two groups corresponding to two different daemon installations and now decided to run just one group, I should have the flexibility to remove or disable the other. Lets go ahead and merge this request in. We can track an issue to handle the scenario where groups have changed.

astroseger: ok! so I merge it.

PR #226, opened 2019-03-01
More user friendly extending and adding funds to the channels

astroseger:
This PR does following:

- In all commands which extend expiration of the channel we check that new expiration is not smaller then old one (which caused non-intuitive error before).
- Add 'snet channel add-for-service' which add funds to the channel opened for specific service 
- Add 'snet channel extend-for-service' which extend expiration for the channel opened for specific service

```
# add 0.1 AGI to the channel which was opened for snet/example-service (initialization is not required)
snet channel add-for-service snet example-service 0.1 -yq

# set expiration 10 days in the future (if old expiration was actually bigger will fail gracefully)
snet channel extend-for-service snet example-service +10days -yq

```


astroseger:
I agree with @arturgontijo that we could unify all functions which add/extend channels into one function. I would simply call it `extend-add`. 
```
snet channel extend-add --service snet example-service --amount 0.1 -yq
snet channel extend-add --service snet example-service --expiration +10days -yq
snet channel extend-add --service snet example-service --expiration +10days --amount 0.1 -yq

# I would use --channel-id, not --id because we use channel-id everywhere else. 
snet channel extend-add --channel-id 1200 --amount 0.1 -yq
snet channel extend-add --channel-id 1200 --expiration +10days -yq
snet channel extend-add --channel-id 1200 --amount 0.1 --expiration +10days -yq

# you cannot use --channel-id  and --service together, so you ether specify the service or your directly set the channel-id
```

But it is API change (we change signature of 'extend-add' and remove 'extend-add-for-service'). I think that it is ok to do at this stage. 
@raamb  do your agree?

raamb:
@astroseger can you please elaborate on what you mean by API change. The signature change of extend-add should not impact any external component correct?

astroseger:
@raamb 
As other components I meant dev-portal (tutorial and examples) :) 
So if there is no objections I will fuse everything into ```extend-add```

raamb:
Got it. In that can can I request you to please update the tutorials and examples. We can merge them all together. The reason I insist on this is that with a beta launch done we will start encouraging people to play with all the components of the platform and dont want them to run into issues due to our documentation being outdated. Thanks

astroseger:
@arturgontijo @raamb 
For now, I've implemented intermediate solution (simply allow specify only `--amount` or only `--expiration`):
```
snet channel extend-add-for-service snet example-service --amount 0.1 -yq
snet channel extend-add-for-service snet example-service --expiration +10days -yq
snet channel extend-add-for-service snet example-service --expiration +10days --amount 0.1 -yq

snet channel extend-add  1200 --amount 0.1 -yq
snet channel extend-add  1200 --expiration +10days -yq
snet channel extend-add  1200 --amount 0.1 --expiration +10days -yq
```

Merging functionality of `extend-add` and `extend-add-for-service` would require more work, and I'm not 100% sure that it would be the best solution, because  `extend-add-for-service` and `extend-add` are quite different.

PR #231, opened 2019-03-14
helper command to print contract addresses for a network

prashantramangupta:
helper command to print contract addresses(Registry and MPE)  for current/default network id


prashantramangupta:
@astroseger please review

Issue #193, opened 2019-01-28
Getting the service health/status


glautervl:
It would be useful if we could get the service health/status using the CLI.

astroseger:
fixed by #236

PR #233, opened 2019-03-21
getting the service health/status

prashantramangupta:
getting the service health/status by making the grpc call to the service endpoint using grpc standard proto. Resolves Issue #193

### DIFF
--- a/snet_cli/mpe_service_command.py
+++ b/snet_cli/mpe_service_command.py
@@ -1,10 +1,13 @@
 from snet_cli.commands    import BlockchainCommand
 import snet_cli.utils_ipfs as utils_ipfs
 from snet_cli.mpe_service_metadata import MPEServiceMetadata, load_mpe_service_metadata, mpe_service_metadata_from_json
-from snet_cli.utils import type_converter, bytes32_to_str
+from snet_cli.utils import type_converter, bytes32_to_str, open_grpc_channel
 from snet_cli.utils_ipfs import hash_to_bytesuri, bytesuri_to_hash, get_from_ipfs_and_checkhash, safe_extract_proto_from_ipfs
 import web3
 import json
+import grpc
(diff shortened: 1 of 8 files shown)


snet_cli/mpe_service_command.py
```diff
@@ -171,6 +174,32 @@ def print_service_metadata_from_registry(self):
         metadata      = self._get_service_metadata_from_registry()
         self._printout(metadata.get_json_pretty())
 
+    def _service_status(self, url, secure=True):
+        channel = None
+        try:
+            channel = open_grpc_channel(endpoint=url)
+            stub = heartb_pb2_grpc.HealthStub(channel)
+            try:
```
raamb: Lets add a timeout for the healthcheck as well. We can go with 15 seconds

prashantramangupta: added timeout as 10 sec same as in Dapp

astroseger:
1. How snet-cli will find health_pb2 and health_pb2_grpc? I would propose to put corresponding proto file in ```snet_cli/resources/proto```. In this case it will be compiled automatically and you will be able to import it (as ```snet_cli.resources.proto.<name_of_file>```)
2. It will not work for the service with multiply groups. You should add ```add_p_group_name(p)``` parameter.
3. So health service doesn't return any interesting information, so we can print only Available/Not Available?
4. I would add some CI tests. But it will require add heart beats in our dummy service... We could do it later...

prashantramangupta:
@astroseger
1.  health_pb2 and health_pb2_grpc is part of the standard grpc package(grpcio-health-checking==1.19.0)
2. It is expected to give available/unavailable info. Let me know if you need more info here.
3. working on this 
**It will not work for the service with multiple groups. You should add add_p_group_name(p) parameter.**

prashantramangupta:
Points are noted and work is in progress

Issue #215, opened 2019-02-22
Helper command to print contract addresses for a network


ferrouswheel:
Instead of having to look up documentation it'd be nice to just make it easy to print that information since the snet-cli has that knowledge already.

astroseger:
In ```snet session```?

ferrouswheel:
yeah, that seems like the best place, probably under the network section as `default_registry_address` etc

```
$ snet session
    identity.joel:
        default_wallet_index: '0'
        identity_type: key
    ipfs:
        default_ipfs_endpoint: http://ipfs.singularitynet.io:80
    network.ropsten:
        default_eth_rpc_endpoint: https://ropsten.infura.io
        default_gas_price: '1000000000'
    session:
        identity: joel
        network: ropsten
```

astroseger:
was fixed by #232

PR #235, opened 2019-03-25
Helper command to print contract addresses for a network

prashantramangupta:
helper command to print contract addresses for a network(registry and mpe)

### DIFF
--- a/snet_cli/commands.py
+++ b/snet_cli/commands.py
@@ -12,7 +12,7 @@ from snet_cli.contract import Contract
 from snet_cli.identity import get_kws_for_identity_type
 from snet_cli.utils import DefaultAttributeObject, get_web3, serializable, type_converter, get_contract_def, get_cli_version, bytes32_to_str
 
-from snet_cli.utils_config import get_contract_address, get_field_from_args_or_session
+from snet_cli.utils_config import get_contract_address, get_field_from_args_or_session, read_default_contract_address
 from snet_cli.identity import RpcIdentityProvider, MnemonicIdentityProvider, TrezorIdentityProvider, \
     LedgerIdentityProvider, KeyIdentityProvider, KeyStoreIdentityProvider
 from web3.eth import is_checksum_address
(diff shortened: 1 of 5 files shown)


astroseger:
Thanks for splitting PR (however you shouldn't be afraid to reference issues which PR fix in PR itself #215)

snet_cli/commands.py
```diff
@@ -263,22 +263,40 @@ def set(self):
...
         # we don't want to who private_key and mnemonic
         for d in rez.values():
             d.pop("private_key", None)
             d.pop("mnemonic", None)
         self._pprint(rez)
 
+    def populate_contract_address(self, rez, key):
+        try:
+            rez[key]['default_registry_address'] = read_default_contract_address(w3=self.w3, contract_name="Registry",
+                                                     err_msg="unable to fetch registry address")
```
astroseger: You could leave err_msg empty. You catch Exception in the same function and never use this err_msg!

snet_cli/commands.py
```diff
@@ -263,22 +263,40 @@ def set(self):
...
+            rez[key]['default_multiparty_escrow_address'] = read_default_contract_address(w3=self.w3, contract_name="MultiPartyEscrow",
```
astroseger: For sake of consistency I would use the following names: ```default_multipartyescrow_at``, ```default_registry_at```,  ```default_singularitynettoken_at``` (because we already have ```current_multipartyescrow_at```, ```current_registry_at```, ```current_singularitynettoken_at```).

snet_cli/commands.py
```diff
@@ -263,22 +263,40 @@ def set(self):
...
+                                                     err_msg="unable to fetch mpe address")
+            rez[key]['default_singularity_net_token_address'] = read_default_contract_address(w3=self.w3, contract_name="SingularityNetToken",
+                                                     err_msg="unable to fetch singularity-net-token address")
+        except Exception as e:
+            rez[key]['default_registry_address'] = None
```
astroseger: I would propose to not set default values if we cannot find them.

So I propose to do the following:
```
      except Exception as e:
             pass
     
```

PR #236, opened 2019-03-25
 getting the service health/status

prashantramangupta:
getting the service health/status by making the grpc call to the service endpoint using grpc standard proto. Resolves Issue #193

### DIFF
--- a/snet_cli/mpe_service_command.py
+++ b/snet_cli/mpe_service_command.py
@@ -1,10 +1,13 @@
 from snet_cli.commands    import BlockchainCommand
 import snet_cli.utils_ipfs as utils_ipfs
 from snet_cli.mpe_service_metadata import MPEServiceMetadata, load_mpe_service_metadata, mpe_service_metadata_from_json
-from snet_cli.utils import type_converter, bytes32_to_str
+from snet_cli.utils import type_converter, bytes32_to_str, open_grpc_channel
 from snet_cli.utils_ipfs import hash_to_bytesuri, bytesuri_to_hash, get_from_ipfs_and_checkhash, safe_extract_proto_from_ipfs
 import web3
 import json
+import grpc
+from grpc_health.v1 import health_pb2 as heartb_pb2
+from grpc_health.v1 import health_pb2_grpc as  heartb_pb2_grpc
 
 class MPEServiceCommand(BlockchainCommand):
 
@@ -170,6 +173,32 @@ class MPEServiceCommand(BlockchainCommand):
     def print_service_metadata_from_registry(self):
         metadata      = self._get_service_metadata_from_registry()
         self._printout(metadata.get_json_pretty())
+
+    def _service_status(self, url, secure=True):
+        channel = None
+        try:
+            channel = open_grpc_channel(endpoint=url)
+            stub = heartb_pb2_grpc.HealthStub(channel)
+            try:
+                response = stub.Check(heartb_pb2.HealthCheckRequest(service=""))
+            except grpc.RpcError as err:
+                err_code = str(err.code())
+                print(err_code)
+            if response!=None and response.status == 1:
+                return True
+            return False
+        except Exception as e:
+            print("error in making grpc call::url: ", url, "|err: ", e)
+            return False
+
+    def print_service_status(self):
+        metadata      = self._get_service_metadata_from_registry().m
+        endpoints =[]
+        for endpoint in metadata.get("endpoints", {}):
+            endpoint["status"] = "Available" if self._service_status(url=endpoint['endpoint'])  else "Not Available"
+            endpoints.append(endpoint)
+        self._pprint(endpoints)
+
 
     def print_service_tags_from_registry(self):
         rez  = self._get_service_registration()
--- a/snet_cli/mpe_service_command.py
+++ b/snet_cli/mpe_service_command.py
@@ -180,7 +180,7 @@ class MPEServiceCommand(BlockchainCommand):
             channel = open_grpc_channel(endpoint=url)
             stub = heartb_pb2_grpc.HealthStub(channel)
             try:
-                response = stub.Check(heartb_pb2.HealthCheckRequest(service=""))
+                response = stub.Check(heartb_pb2.HealthCheckRequest(service=""), timeout=10)
             except grpc.RpcError as err:
                 err_code = str(err.code())
                 print(err_code)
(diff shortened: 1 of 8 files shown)


snet_cli/mpe_service_command.py
```diff
@@ -171,6 +174,40 @@ def print_service_metadata_from_registry(self):
         metadata      = self._get_service_metadata_from_registry()
         self._printout(metadata.get_json_pretty())
 
+    def _service_status(self, url, secure=True):
+        channel = None
```
astroseger: Why do you set channel to None?

snet_cli/mpe_service_command.py
```diff
@@ -171,6 +174,40 @@ def print_service_metadata_from_registry(self):
...
+        try:
+            channel = open_grpc_channel(endpoint=url)
+            stub = heartb_pb2_grpc.HealthStub(channel)
+            try:
+                response = stub.Check(heartb_pb2.HealthCheckRequest(service=""), timeout=10)
+            except grpc.RpcError as err:
+                err_code = str(err.code())
+                print(err_code)
```
astroseger: please use ```self._printerr``` or ```self._printout``` instead of `print`. By the way, do you really want to print something or it was a debug message?
Because, If you don't want to print this error messages then you don't need this complicated nested try. You could simply write:
```python
        try:
            channel = open_grpc_channel(endpoint=url)
            stub = heartb_pb2_grpc.HealthStub(channel)
            response = stub.Check(heartb_pb2.HealthCheckRequest(service=""), timeout=10)
            if response!=None and response.status == 1:
                return True
        except Exception as e:
                return False
```

snet_cli/mpe_service_command.py
```diff
@@ -171,6 +174,40 @@ def print_service_metadata_from_registry(self):
...
+            if response!=None and response.status == 1:
+                return True
+            return False
+        except Exception as e:
+            print("error in making grpc call::url: ", url, "|err: ", e)
+            return False
+
+    def print_service_status(self):
+        metadata      = self._get_service_metadata_from_registry().m
```
astroseger: Please don't use internal structure of service_metadata. Too bad python doesn't have private variables... If you need some functionality from ```MPEServiceMetadata``` which it doesn't have, please add this functionality inside MPEServiceMetadata. 

Which functionality do we want in case of multiply payment groups? I see two options.
1. Force user to obligatory specify ```group_name``` in case of multiply group names, as we do in the rest of the code. 
In this case you simply could write:
```python
        metadata      = self._get_service_metadata_from_registry()
        endpoints = metadata.get_endpoints_for_group(self.args.group_name)
        for endpoint in endpoints:
            status = "Available" if self._service_status(url=endpoint)  else "Not Available"
            self._printout("%s   %s"%(endpoint, status))
```

2.  Print status for all endpoints in all groups. Now I understand that your initial code did something like this. Sorry that I didn't realize it initially. You only should use ```get_all_endpoints``` function which we already have in metadata class. And for course you don't need group_name parameter in this case.
```python
        metadata      = self._get_service_metadata_from_registry()
        endpoints = metadata.get_all_endpoints()
        for endpoint in endpoints:
            status = "Available" if self._service_status(url=endpoint)  else "Not Available"
            self._printout("%s   %s"%(endpoint, status))
```

I think that both solutions are ok...

astroseger:
I would suggest to add tests if it is not too complicated. To make it absolutely clean, It would be perfect to add heart beat to our dummy service ```snet-cli/test/functional_tests/simple_daemon/test_simple_daemon.py```. But actually it is not necessary, because our command must work even if service does not implement heart beat.

Our plan minimum is following:
For example in ```script13_call_reinitialize.sh```:
- after ```snet service publish testo tests -y -q``` (line 14) add ```snet service print-service-status testo tests``` 
- after  ```snet service publish testo tests2 -y -q``` (line 57) add ```snet service print-service-status testo tests2```

You also can make separate test scenario for ```print-service-status``` (for example ```script14_print_service_status.sh```), but I don't think it is necessary and it is ok to add calls in existed scenario.

snet_cli/mpe_service_metadata.py
```diff
@@ -171,6 +171,14 @@ def get_payment_address(self, group_name = None):
     def get_all_endpoints(self):
         return [e["endpoint"] for e in self.m["endpoints"]]
 
+    def get_all_endpoints_with_group_name(self):
+        endpts_with_grp = {}
```
astroseger: You could use defaultdict (defaultdict(list)) to simplify code:
```python
from collections import defaultdict 
....
def get_all_endpoints_with_group_name(self):
        endpts_with_grp = defaultdict(list)
        for e in self.m["endpoints"]:
            endpts_with_grp[e['group_name']].append(e['endpoint'])
        return endpts_with_grp
```

snet_cli/mpe_service_metadata.py
```diff
@@ -171,6 +171,14 @@ def get_payment_address(self, group_name = None):
...
+        for e in self.m["endpoints"]:
+            if e['group_name'] not in endpts_with_grp.keys():
```
astroseger: you don't need ```.keys()``` here ```x in dict``` is equal to ```x in dict.keys()```  (but I would use defaultdict here to simplify the code, as I've said in my first comment)

snet_cli/mpe_service_command.py
```diff
@@ -171,6 +174,36 @@ def print_service_metadata_from_registry(self):
...
+        else:
+            groups = metadata.get_all_endpoints_with_group_name()
+        srvc_status = {}
+        for grp in groups:
+            srvc_status[grp] = {}
+            if groups[grp] == []:
+               srvc_status.pop(grp)
+            for endpoint in groups[grp]:
+                status = "Available" if self._service_status(url=endpoint)  else "Not Available"
+                srvc_status[grp].update({"endpoint": endpoint, "status": status})
```
astroseger: update? It will not work for several endpoints because you will print information only about the last endpoint.

I would propose to rewrite like this (which also simplify code):
```python
        srvc_status = defaultdict(list)
        for grp in groups:
            for endpoint in groups[grp]:
                status = "Available" if self._service_status(url=endpoint)  else "Not Available"
                srvc_status[grp].append({"endpoint": endpoint, "status": status})
        if srvc_status == {}:
            self._printout("Error: No endpoints found to check service status.")
            return
        self._pprint(srvc_status)

```

snet_cli/mpe_service_metadata.py
prashantramangupta: noted and updated. Will make use of defaultdict, it makes the code simpler.

Issue #196, opened 2019-01-30
Change version number when something changes


ramongduraes:
Running `snet version` is the best way to keep track of our installed version vs the latest release. However, SNET CLI has been changing quite a bit but the version numbers do not indicate that. 

For instance, my colleagues and I all had snet-cli version 0.1.9 installed but only one of us had `snet service metadata-set-model` command available. After pulling the repository and reinstalling it everything worked. Since we're using the MAJOR.MINOR.PATCH versioning system, the patch number could've changed to 0.1.10 to indicate that something has changed.

ramongduraes:
Adding build number to `snet version` is also an option

ferrouswheel:
The MAJOR.MINOR.PATCH of `snet version` shouldn't change until there is a packaged release.

However automatically adding the git hash to versions installed directly from the repo would be a sensible way to indicate development versions are different.

prashantramangupta:
@astroseger @raamb 
1. we should have version.py in our package, from where we will read the version in setup.py. Currently, we are reading from __init__.py
2. code build will write a version in version.py during the build
https://stackoverflow.com/questions/2058802/how-can-i-get-the-version-defined-in-setup-py-setuptools-in-my-package

raamb:
@vinthedark now that pr #237 has been merged, lets update the client CD to set the version number in the build, similar to what we did on the daemon.

vinthedark:
Setting up the version number as a part of the CD is been taken care. Going forward, the version number will be updated during every github release.

PR #237, opened 2019-03-25
keeping version details in standard version.py file

prashantramangupta:
Resolves issue #196
1. We are removing __version__ key from __init__.py and keeping it in separate file version.py.
2. The code build process (managed in AWS) will update the version in version.py on every release.


raamb:
> The code release will update the `version.py` here on Github too?

Yes, we will have code build update the version in Github. Since that's an external script wanted to restrict it to a separate file.

> I did this (put it inside `__init__.py`) following modules like [Flask/setup.py](https://github.com/pallets/flask/blob/master/setup.py) and [AWS-CLI/setup.py](https://github.com/aws/aws-cli/blob/develop/setup.py)
> 
> I'd like to suggest to keep it like it is and just increment the version when we improve something that worth it.
Refer to https://packaging.python.org/guides/single-sourcing-package-version/, technique 6 refers to the approach that you mention but it has a potential downside. Keeping the version in its own file makes it simple and easy for the build process to update it during every release (techniques 1 thru 5 are on these lines)

arturgontijo:
> Yes, we will have code build update the version in Github. Since that's an external script wanted to restrict it to a separate file.

So it makes sense.
I'd like to suggest 2 things:

1 - Move `version.py` to the package `snet_cli` folder. (in `setup.py` we can use `from snet_cli.version import __version__`)
2 - Import it in `__init__.py` too, so the users can get the version like:
```
>>> import snet_cli
>>> snet_cli.__version__
'0.2.10'
```

setup.py
```diff
@@ -15,21 +14,27 @@ def install_and_compile_proto():
...
     """Post-installation for installation mode."""
+
     def run(self):
         _install.run(self)
         self.execute(install_and_compile_proto, (), msg="Compile protocol buffers")
 
 
-with open('snet_cli/__init__.py', 'rt', encoding='utf8') as f:
-    version = re.search(r'__version__ = "(.*?)"', f.read()).group(1)
+version = {}
```
astroseger: 1. What is the reason to set version to ```{}``` at line 34???
2. Is using ```exec``` is necessary here? We could use ```import``` here.

```
import snet_cli.version
version = snet_cli.version.__version__
```

astroseger: Sorry... There is an error in my previous comment. Of course you simple should do as following:

```python
from snet_cli.version import __version__ as version
```

prashantramangupta: with
`from snet_cli.version import __version__ as version`
1. This from statement causes the ./snet_cli/__init__.py to be invoked
2. This causes a build failure as some of the dependencies would not have been set up when setup.py is run. This caused the circleci build to fail

prashantramangupta: @astroseger Can you approve the pr if you are ok?

arturgontijo:
Why we can't have this?

> 2 - Import it in `__init__.py` too, so the users can get the version like:
> ```
> >>> import snet_cli
> >>> snet_cli.__version__
> '0.2.10'
> ```

raamb:
@arturgontijo apologies missed your comment. We can do this, but is there any reason to have it?

arturgontijo:
It is a good practice for python modules:

OpenCV
```
>>> import cv2
>>> cv2.__version__
'3.4.2'
```

AWS-Cli
```
>>> import awscli
>>> awscli.__version__
'1.16.89'
```

But it's just a suggestion, it isn't so important.

PR #238, opened 2019-03-29
[Arguments] Alphabetical Order and Uppercase Arguments.

arturgontijo:
- Tried to sort all commands and subcommands in alphabetical order.
- Using uppercase in all arguments.
- Pycharm has changed some text parts of the code (like adding newlines). I think this way the code got cleaner, but I can change it back.


arturgontijo:
CI is stuck.

PR #239, opened 2019-03-29
[Arguments] Alphabetical Order and Uppercase Arguments.

arturgontijo:
- Tried to sort all commands and subcommands in alphabetical order.
- Using uppercase in all arguments.
- Pycharm has changed some text parts of the code (like adding newlines). I think this way the code got cleaner, but I can change it back.

### DIFF
--- a/snet_cli/arguments.py
+++ b/snet_cli/arguments.py
@@ -4,37 +4,39 @@ import re
 import sys
 from pathlib import Path
 
-from snet_cli.commands import IdentityCommand, SessionSetCommand, SessionShowCommand, NetworkCommand, ContractCommand, OrganizationCommand, VersionCommand
+from snet_cli.commands import IdentityCommand, SessionSetCommand, SessionShowCommand, NetworkCommand, ContractCommand, \
+    OrganizationCommand, VersionCommand
 from snet_cli.identity import get_identity_types
 from snet_cli.utils import type_converter, get_contract_def
 from snet_cli.mpe_account_command import MPEAccountCommand
 from snet_cli.mpe_service_command import MPEServiceCommand
 from snet_cli.mpe_channel_command import MPEChannelCommand
-from snet_cli.mpe_client_command  import MPEClientCommand
+from snet_cli.mpe_client_command import MPEClientCommand
 from snet_cli.mpe_treasurer_command import MPETreasurerCommand
 from snet_cli.sdk_command import SDKCommand
 from snet_cli.utils_agi2cogs import stragi2cogs
 from snet_cli.config import get_session_keys, get_session_network_keys_removable
 
+
 class CustomParser(argparse.ArgumentParser):
     def __init__(self, default_choice=None, *args, **kwargs):
         self.default_choice = default_choice
         super().__init__(*args, **kwargs)
-
+    
     def error(self, message):
         sys.stderr.write("error: {}\n\n".format(message))
         self.print_help(sys.stderr)
         sys.exit(2)
-
+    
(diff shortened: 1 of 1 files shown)


astroseger:
@arturgontijo  
Hmm.... Why do you want to sort command alphabetically? We have quite a lot commands in ```snet service``` and ```snet channel``` and before they where sorted kind of by functionality which I think helped a little bit, but with alphabetical order everything is mixed up....

arturgontijo:
I did it because while I was updating the docs I checked that the order was not that good (of the first layer commands).
But reading my modifications now, I agree that the sub-commands will be better presented in a logical (workflow) order.
What do you think about sort only the first layer commands?

astroseger:
Yes we certainly should sort the first layer commands.
For second layer commands: we probably could group them into groups and sort them inside groups, or we could leave them as it is.

arturgontijo:
The group idea is great but I think that that the subcommands naming structure are already grouping them.

What do you think about this new structure? (* = modification)

```
snet account [-h] COMMAND ...
			print
			balance
			deposit
			withdraw
			transfer

snet channel [-h] COMMAND ...
			init
			init-metadata
			open-init
			open-init-metadata
			extend-add
			extend-add-for-service
			block-number
			print-initialized
			print-initialized-filter-service
			print-all-filter-sender
			print-all-filter-recipient
			print-all-filter-group
			print-all-filter-group-sender
			claim-timeout *
			claim-timeout-all *

snet client [-h] COMMAND ...
			call
			call-lowlevel
			get-channel-state

snet contract [-h] CONTRACT ...
			MultiPartyEscrow
			Registry
			SingularityNetToken

snet identity [-h] ACTION ...
			list
			create
			delete
			ID_0
			ID_1  
			ID_2

snet network [-h] NETWORK ...
			list
			create
			kovan
			mainnet
			ropsten
			rinkeby

snet organization [-h] COMMAND ...
			list
			list-org-names
			list-my
			list-services *
			info
			create
			change-name
			change-owner
			add-members
			rem-members
			delete *

snet sdk [-h] COMMAND ...
			generate-client-library

snet service [-h] COMMAND ...
			metadata-init
			metadata-set-model
			metadata-set-fixed-price
			metadata-add-group
			metadata-add-endpoints
			metadata-remove-all-endpoints
			metadata-update-endpoints
			metadata-add-description
			publish
			publish-in-ipfs *
			update-metadata
			update-add-tags
			update-remove-tags
			print-metadata
			print-service-status
			print-tags
			get-api-metadata
			get-api-registry
			delete

snet session [-h]

snet set [-h] KEY VALUE
			KEY
			VALUE

snet treasurer [-h] COMMAND ...
			print-unclaimed
			claim
			claim-all
			claim-expired

snet unset [-h] KEY
			KEY

snet version [-h]
```

astroseger:
looks good for me

snet_cli/arguments.py
```diff
@@ -4,95 +4,113 @@
...
             for action in list(filter(
-                lambda subparser_action: isinstance(subparser_action, argparse._SubParsersAction),
-                self._subparsers._actions
+                    lambda subparser_action: isinstance(subparser_action, argparse._SubParsersAction),
+                    self._subparsers._actions
             )):
                 if not len(list(filter(lambda arg: arg in action._name_parser_map.keys(), arg_strings))):
                     arg_strings = [self.default_choice] + arg_strings
-
+
```
astroseger: I would remove such white-spaces in empty string (```sed -i 's/[ \t]*$//' arguments.py```), but of course it is very minor..

arturgontijo: Done.

PR #240, opened 2019-04-01
Auto generate commands documentation 

raamb:
Auto generate documentation for snet-cli commands using Sphinx argparse. 
~~Currently the top level sub commands are coded into the index.rst file, in the next iteration we will attempt to automate this as well.~~

We now auto generate documentation. Wrote up a simple python script which 
- Gets the parser object and obtains all the top level sub commands
- For each subcommand check if its rst file is present else generate it
- Add all the subcommands to the index.rst in a sorted fashion
- Build html docs using sphinx.

The main reason I am doing above is to organize the commands better.


astroseger:
1. From the point of view of functionality everything is working. I was able to generate documentation.
2. Do we need ``` docs/make.bat```? I don't think that we will need to generate documentation under windows.
3. It seems ```docs/source/snet-cli-static/theme.css``` was somehow generated using sphinx-rtd-theme. Automatic generation of this file is impossible/not practical?
4. I get this warning (but I guess nothing serious...): 
```
writing additional pages... search/usr/local/lib/python3.6/dist-packages/sphinx_rtd_theme/search.html:20: RemovedInSphinx30Warning: To modify script_files in the theme is deprecated. Please insert a <script> tag directly in your theme instead.
```

raamb:
> 2\. Do we need ` docs/make.bat`? I don't think that we will need to generate documentation under windows.

I left it in there for anybody using a windows system. We can remove it

> 3\. It seems `docs/source/snet-cli-static/theme.css` was somehow generated using sphinx-rtd-theme. Automatic generation of this file is impossible/not practical?

So Sphinx auto generates the CSS depending on the theme, I tried overriding specific classes but that didnt work. So I just had customized the generated CSS and checked it in. This gives us a SingularityNet look and feel (to some extent). I was unable to crack auto generating the CSS with the changes we needed.

> 4\. I get this warning (but I guess nothing serious...):

We use "Read the docs theme" which is generating this warning. We can ignore for now.

raamb:
Done. Removed the .bat file.

PR #241, opened 2019-04-04
Full stateless logic

astroseger:
Add full support of stateless logic in snet-cli (see https://github.com/singnet/snet-daemon/issues/220 and  https://github.com/singnet/dev-portal/blob/master/docs/concepts/mpe-stateless-client.md).

With full support of stateless logic we can calculate unspent amount in situation when ```current_nonce != blockchain_nonce```.  It can happen after service provider initiate claim procedure, but blockchain claim transaction hasn't been claimed yet (or this blockchain transaction was canceled for some reason). You can simulate this situation by running ```snet treasurer claim-all ...``` and cancel blockchain transaction (press ```n``` or directly run ```echo n | snet treasurer claim-all ... ```).

This PR does following:
1. Add full support of stateless logic. If daemon implement https://github.com/singnet/snet-daemon/issues/220 then  ```snet client get-channel-state``` will return correct ```unspent_amount_in_cogs``` in all situation (even when ```current_nonce != blockchain_nonce```)
2. Add full support of stateless logic in our dummy test daemon
3. Add functional tests for this stateless logic  (```script14_full_stateless_logic.sh```)

It should be noted that the only one function in snet-cli which is affected by this PR is ```snet client get-channel-state```.

This PR should be backward compatible with old daemon!


astroseger:
@raamb @arturgontijo @Dinesh2521 @vsbogd 
Please review this PR. 
Daemon now support full stateless logic, so we could merge this PR. I've checked it, and it is working with new daemon.

arturgontijo:
Hi @astroseger , sorry about the delay...
Starting the review I've found this (after `pip install -e .` and `./scripts/blockchain install`):
```
$  snet account balance
Error: when sending a str, it must be a hex string. Got: '0x0xff2a327ed1ca40ce93f116c5d6646b56991c0dde'
If you want to see full Traceback then run:
snet --print-traceback [parameters]
```
I think this was discussed in another PR. Maybe you must update your branch code.

astroseger:
> Hi @astroseger , sorry about the delay...
> Starting the review I've found this (after `pip install -e .` and `./scripts/blockchain install`):
> 
> ```
> $  snet account balance
> Error: when sending a str, it must be a hex string. Got: '0x0xff2a327ed1ca40ce93f116c5d6646b56991c0dde'
> If you want to see full Traceback then run:
> snet --print-traceback [parameters]
> ```
> 
> I think this was discussed in another PR. Maybe you must update your branch code.

Please try again. I've pulled master where it has been already fixed.

snet_cli/mpe_client_command.py
```diff
@@ -149,14 +157,10 @@ def call_server_lowlevel(self):
...
     def _get_channel_state_from_server(self, grpc_channel, channel_id):
-        # Compile protobuf if needed
-        codegen_dir = Path.home().joinpath(".snet", "mpe_client", "state_service")
-        proto_dir   = Path(__file__).absolute().parent.joinpath("resources", "proto")
-        if (not codegen_dir.joinpath("state_service_pb2.py").is_file()):
-            compile_proto(proto_dir, codegen_dir, proto_file = "state_service.proto")
-
-        # make PaymentChannelStateService.GetChannelState call to the daemon
-        stub_class, request_class, _ = import_protobuf_from_dir(codegen_dir, "GetChannelState")
+        # We should simply statically import everything, but it doesn't work because of the following issue in protobuf: https://github.com/protocolbuffers/protobuf/issues/1491
```
arturgontijo: Is it worth to add `resources/proto` to `sys.path`?
```
proto_dir = Path(__file__).absolute().parent.joinpath("resources", "proto")
if str(proto_dir) not in sys.path:
    sys.path.append(str(proto_dir))

from state_service_pb2 import ChannelStateRequest as request_class
from state_service_pb2_grpc import PaymentChannelStateServiceStub as stub_class
```
But I'm not sure that this way would be better than yours (only if this can be re-use in multiple parts)

snet_cli/mpe_client_command.py
```diff
@@ -9,7 +9,6 @@
 from snet_cli.utils_agi2cogs import cogs2stragi
 from snet_cli.utils import open_grpc_channel, rgetattr, compile_proto
```
arturgontijo: Unused imports:
- import grpc
- compile_proto

snet_cli/mpe_client_command.py
```diff
@@ -30,6 +29,11 @@ def _verify_my_signature(self, signature, mpe_address, channel_id, nonce, amount
         sign_address = self.ident.w3.eth.account.recoverHash(message_hash, signature=signature)
         return sign_address == self.ident.address
 
+    def _assert_vilidity_of_my_signature_or_zero_amount(self, signature, channel_id, nonce, signed_amount, error_message):
```
arturgontijo: `_assert_vilidity_of_my_signature_or_zero_amount` -> `_assert_validity_of_my_signature_or_zero_amount`
or maybe, just:
`_assert_vilidity_of_my_signature_or_zero_amount` -> `_assert_signature_or_zero_amount`

snet_cli/mpe_client_command.py
astroseger: Let's consider it later when this story with monorepo will end.

astroseger:
I've switched to new authentication in GetChannelState (As was agreed with @anandrgitnirman ). However https://github.com/singnet/snet-daemon/pull/285 should be merged into daemon first.

astroseger:
@raamb 
I've resolved marge conflict.

PR #242, opened 2019-04-08
Support saving nested fields from service response

ferrouswheel:
`--save-field` currently assumes that protobuf messages are flat, this PR allows the user to use a dot notation to access nested fields.

Example use:

```
snet channel open-init snet semantic-segmentation  0.001 +10days
snet client call --save-field debug_img.content output.jpg snet semantic-segmentation segment '{"img": {"file@content":"test.jpg", "mimetype": "image/jpeg"}, "visualise":true}'
```

### DIFF
--- a/snet_cli/mpe_client_command.py
+++ b/snet_cli/mpe_client_command.py
@@ -113,12 +113,24 @@ class MPEClientCommand(MPEChannelCommand):
                     ("snet-payment-channel-signature-bin", bytes(signature))]
         return call_fn(request, metadata=metadata)
 
+    def _get_field(self, field, response):
+        field_parts = field.split(".", 1)
+        field_name = field_parts[0]
+        try:
+            field_value = getattr(response, field_name)
+            if len(field_parts) == 2:
+                return self._get_field(field_parts[1], field_value)
+            else:
+                return field_value
+        except AttributeError:
+            raise Exception("Field {} not found in message" % field_name)
+
     def _deal_with_call_response(self, response):
         if (self.args.save_response):
             with open(self.args.save_response, "wb") as f:
                 f.write(response.SerializeToString())
         elif (self.args.save_field):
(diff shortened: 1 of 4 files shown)


astroseger:
Actually there is even shorter way (because we don't "default" functionality of getattr)
```python
def rgetattr(obj, attr):
    return functools.reduce(getattr, [obj] + attr.split('.'))
```

ferrouswheel:
Updated to using `functools`

I also changed circleci to executing the unit tests using `nosetests`, since we were ignoring doctests.

snet_cli/mpe_client_command.py
```diff
@@ -118,7 +117,8 @@ def _deal_with_call_response(self, response):
             with open(self.args.save_response, "wb") as f:
                 f.write(response.SerializeToString())
         elif (self.args.save_field):
-            field = getattr(response, self.args.save_field[0])
+            field = rgetattr(response, self.args.save_field[0])
+            #field = self._get_field(, response)
```
astroseger: You've forget to remove this comment :)

test/utils/run_all_unit.sh
```diff
@@ -1,4 +1 @@
-for f in test/unit_tests/*.py
-do
-   bash -ex -c "cd test/unit_tests; python `basename $f`"
-done
+nosetests -v --with-doctest
```
astroseger: We probably don't need run_all_utils.sh anymore...

snet_cli/mpe_client_command.py
ferrouswheel: doh- removed it

test/utils/run_all_unit.sh
ferrouswheel: I have removed the file.

ferrouswheel:
Same random fail as the other PR had #244

PR #246, opened 2019-04-10
New sdk witch comply to new sdk-API (temporaly solution)

astroseger:
Temporally version of SDK which comply to new API (see https://github.com/singnet/snet-sdk-python/issues/16).
This version will be replaced by more clean version by @vforvalerio87 which will add one level of abstraction between SDK and "Commands".  But API will not change so you can use this SDK as it is..

This PR adds:
* new SDK which comply to new API
* CI for this SDK
* SDK supports json encoding
* Two funding strategy (Basic and Automatic)
* Two ways of getting request_class (see example and discussion in https://github.com/singnet/snet-sdk-python/issues/16)

Simple example
```python
from snet_cli.sdk import Session, Client, AutoFundingFundingStrategy
from snet_cli.config import Config

# we directly use snet-cli configuration
# another types of configuration can be easiliy implemented (for example hard-coded configuration)
session = Session(Config())

# In case of multiply grpc_service names you should specify the name of grpc_service
# Here we have only one grpc_service (Calculator), so we can omit this parameter
client = Client(session, "snet", "example-service", AutoFundingFundingStrategy(amount_cogs = 4, expiration = "+10days"))
for i in range(10):
    request = client.classes.Numbers(a=20, b=i)
    rez = client.stub.add(request)
    print("add", rez)
    request = client.get_request_class("mul")(a=20, b=i)
    rez = client.stub.mul(request)
    print("mul", rez)
```


vsbogd:
Sergey, could you please rebase your branch on latest master

PR #253, opened 2019-05-16
PEP8 guidelines (first attempt).

arturgontijo:
First modifications toward full PEP8 recommendations (including the < 80 chars per line).
To get 100% coverage we'll need to change some functions and variables names after this PR.

### DIFF
--- a/scripts/blockchain
+++ b/scripts/blockchain
@@ -8,40 +8,65 @@ import sys
 
 
 def main():
-    assert len(sys.argv) > 1, "please select a target from 'install', 'uninstall'"
+    assert (len(sys.argv) > 1,
+            "please select a target from 'install', 'uninstall'")
     target = sys.argv[1]
-    blockchain_dir = pathlib.Path(__file__).absolute().parent.parent.joinpath("blockchain")
+    package_folder = pathlib.Path(__file__).absolute().parent.parent
+    blockchain_dir = package_folder.joinpath("blockchain")
     node_modules_dir = blockchain_dir.joinpath("node_modules")
-    platform_json_src_dir = node_modules_dir.joinpath("singularitynet-platform-contracts")
-    token_json_src_dir = node_modules_dir.joinpath("singularitynet-token-contracts")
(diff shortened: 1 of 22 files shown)


astroseger:
ok 80 lines contradicts to my habit to give self explanatory names to variables/functions... Which I believe is right. And this 80 lines makes code slightly uglier.. 
for example:
```python
channel_blockchain = \
                        self._get_channel_state_from_blockchain(channel_id)
                    value_agi = cogs2stragi(channel_blockchain["value"])
```
or even
```python
raise Exception(
                        "\n\nYou are trying to change group_id "
                        "or payment_address in group '%s'.\n" %
                        old_group["group_name"] +
                        "It would make all open "
                        "channels invalid.\n" +
                        "You shoudn't use 'metadata-init' for "
                        "existed service, because it "
                        "reinitialize group_id\n" +
                        "Please use 'metadata-set-model' "
                        "for change your protobuf file\n" +
                        "Please use "
                        "'metadata-remove-all-endpoints/metadata-add-endpoints"
                        " to update endpoints'\n\n")

```

Please see the code, you can find hundreds of ugly "new lines"..

I believe 80 character requirement is slightly outdated and I wouldn't follow it... 

@vsbogd What do you think?

vsbogd:
Well, I would say I don't have strong opinion on "80 characters per line" rule. Personally I format code to obey it. And I agree with @astroseger that making variable names shorter just to fit in 80 characters per line is not correct way of doing things. 

My argument for it is that there are usually two reasons of failure to apply this rule:
- two many nested blocks of statements; usually it can be fixed by adding smaller methods;
- two many arguments of function; which is also frequently sign of problem in function design.

Anyway I as I said I have never thought that we should strongly stick to "80 characters per line" rule :-)

arturgontijo:
I like to see PEP8 codes but for `snet-cli`, right now, it really messed with the structure a lot.
To make it cleaner we'll have to spend some time on this. =/
I'll close this PR and open a new one without trimming lines.

PR #254, opened 2019-05-20
Monorepo

vforvalerio87:
Set up monorepo:

this monorepo is organized as a PEP 420 compliant namespace
(https://www.python.org/dev/peps/pep-0420/).

The sdk package has been restructured to conform to the PEP 420
namespace package specification and it's under the "snet" namespace.

The cli package is both a regular package and a namespace package.

setup.py for the monorepo itself installs the dependencies for all
packages in the repository.

Each package and the overarching monorepo also have "requirements.txt"
files for local installations and CI enviroments: by installing via
"requirements.txt", the dependencies listed in "setup.py" are only
pulled and installed from PyPI if a local version is not already
available. This allows to get dependencies within the monorepo to be
installed from the local filesystem.

The whole package is called "snet" and gives the name to the namespace.
The subpackages are "snet.sdk", "snet.snet_cli" and snet_cli (since the
CLI package is both within the namespace and a regular package).
Going forward, all of the code in the "snet_cli" package can be migrated
to the namespace package: both were kept momentarily for compatibility.

In this scenario, everything inside the "snet.sdk" package has
visibility on the contents of the "snet.snet_cli" package and
vice-versa.
Everything inside the "snet_cli" regular package has visibility on the
contents of the "snet.snet_cli" package due to the regular Python import
mechanics, but it has no visibility on the contents of the "snet.sdk"
package (and vice-versa).
So, the shared code between the SDK and the CLI goes under the
"snet.snet_cli" package which is at "packages/snet_cli/snet/snet_cli".

Code and assets deduplication:

thanks to the new structure, everything under "resources" has been moved
to the "snet.snet_cli" package and is now shared between the SDK and
CLI. Specifically, this includes the json artifacts from the compilation
and deployment of the smart contracts, the raw ".proto" files to
interact with the daemon and the compiled python client libraries.

Changes:

since the CLI is both a namespace package and a regular package, changes
to the code have been kept to a minimum.
The only change of note is that state service does not compile its own
".proto" files at runtime inside the user's home directory anymore but
they are imported from the "resources" directory under the
"snet.snet_cli" namespace package, where they are compiled during setup.


astroseger:
1. Could you add CI for your SDK (in functional tests). You can take it from: https://github.com/singnet/snet-cli/pull/246
2. I would propose to remove compiled proto files from github ( snet_sdk/resources/proto).

astroseger:
3. Could you give the simple example of usage of your SDK (similar to the example in the description of https://github.com/singnet/snet-cli/pull/246)?

vforvalerio87:
> 1. Could you give the simple example of usage of your SDK (similar to the example in the description of #246)?

https://github.com/singnet/snet-code-examples/pull/3/files

Matches the JS SDK example

astroseger:
> > 1. Could you give the simple example of usage of your SDK (similar to the example in the description of #246)?
> 
> https://github.com/singnet/snet-code-examples/pull/3/files
> 
> Matches the JS SDK example

I've tried to run JS example, but it fails in python ;)

astroseger:
I would propose to not make separate directory for snet_cli and basically do not touch old directory structure for now and just add sdk directory. I afraid this move would make more harm then good (for example CI is brorken because of this now... )

astroseger:
@vforvalerio87 Could you make your PR green before we review it? 
I suggest doing: https://github.com/singnet/snet-cli/pull/254#issuecomment-494343418

If you add some tests for SDK itself then it would be even greater...

raamb:
> I would propose to not make separate directory for snet_cli and basically do not touch old directory structure for now and just add sdk directory. I afraid this move would make more harm then good (for example CI is brorken because of this now... )

I suggest that we fix the CI for the folder structure ASAP. Infact I would propose that we merge in the new folder structure first to ensure that the client functionality is all fine. We can continue to work on the SDK in parallel. @vforvalerio87 lets get the CI for snet-cli fixed right away

astroseger:
@raamb  I do insist that we shoudn't touch directory structure of snet-cli. We could add sdk as separate folder, or, at first directly in snet-cli....
I do not see the reason to move files around... CI is broken because of it, and it is only the beginning of problems this movement will cause.

vforvalerio87:
So, CircleCI config and scripts have been adapted to work in the monorepo scenario, build tests pass.

I tried building and using both the CLI and SDK locally and they worked correctly, feel free to try the same.

I changed the name of the "stub" object to "service" as we discussed.

astroseger:
@vforvalerio87 
Please remove snet_sdk/snet_sdk/state_service_pb2.py and snet_sdk/snet_sdk/state_service_pb2_grpc.py from repo (you can compile them in setup.py as we do in snet-cli)

will you add tests for SDK in this PR?

README.md
```diff
@@ -1,124 +1,18 @@
...
-SingularityNET CLI
+SingularityNET Python Monorepo
 
-## Getting Started  
+## Packages  
+This repository is a monorepo that includes several packages that we publish to PyPI from a shared codebase, specifically:
   
-These instructions are for the development and use of the SingularityNET CLI.
-For further details, please check our full [Documentation](http://snet-cli-docs.singularitynet.io/).
+|Package                                       |Description                                                          |
```
raamb: Lets also link the readme of the cli and SDK here, makes it easier for people to navigate

snet_sdk/scripts/blockchain
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
```
raamb: This is the same file as snet_cli/scripts/blockchain. Why not just reuse it

snet_sdk/snet_sdk/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+from web3.utils.datastructures import AttributeDict, MutableAttributeDict
+
+import snet_sdk.generic_client_interceptor as generic_client_interceptor
+
+from snet_sdk.utils import get_contract_object
+from snet_sdk.service_client import ServiceClient
+from snet_sdk.account import Account
+from snet_sdk.mpe_contract import MPEContract
+from snet_sdk.payment_channel_management_strategies.default import PaymentChannelManagementStrategy
+
```
raamb: Lets use a similar approach to version as we did for the snet-cli i.e. use a version.py file which is changed via CD script. Makes it easier to control versioning in the code and Pypi. Can be taken up as its own PR

vforvalerio87: Sure, I had it like that because it used to be in __init__ when I worked on the first iteration of the SDK... didn't notice it had changed places since then. I'll move it

astroseger:
@vforvalerio87 
* snet_sdk require ```grpcio-tools==1.17.1``` and web3==4.2.1
* snet_cli require ```web3==4.8.3``` and 'grpcio-tools==1.19.0' 
So I was not able to install snet_sdk. After I had changed grpcio-tools and web3 requirement in snet_sdk/setup.py I still was not able to run your example: ```https://github.com/singnet/snet-code-examples/pull/3/files```

So I was not able to run your example...

astroseger:
@vforvalerio87 
I have two additional comments and one request:

Could you provide two examples (better add them directly in functional tests):
1. Example with running "BasicStrategy" which do not extend or add funds to the channel
2. Example with running AutoFundingFundingStrategy with explicitly passing "amount_cogs" + "expiration" (or how you call them :) ).

Comment 1:
We should use the same strategy for funding channels across all components. We've already discussed it with @raamb (about dApp) and we've agreed with the following strategy:
- We take appropriate channel with smallest channelId. 
It means that if there are several channels which were found we simply get channel with smaller channelId.
We do not search for "funded" channel and we do not propose user to open new channel. We always reuse old channel. We are doing it in snet-cli and dApp, so please do the same in SDK.

You are not doing it ```_get_funded_channel``` (you should simply take channel with smallest channelId and return it, if it is not founded)

Comment 2:
You use ```self.metadata.pricing["price_in_cogs"]``` directly which you shouldn't do. We will add new pricing models soon, so we might not have price_in_cogs. I would propose to have a special function  ```get_price_for_next_call```

snet_sdk/snet_sdk/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+        self.account = Account(self.web3, config, self.mpe_contract)
+
+
+    def create_service_client(self, org_id, service_id, service_stub, group_name="default_group", payment_channel_management_strategy=PaymentChannelManagementStrategy, options={}):
+        service_metadata = self.service_metadata(org_id, service_id)
+        try:
+            group = next(filter(lambda group: group["group_name"] == group_name, service_metadata.groups))
+        except StopIteration:
+            raise ValueError("Group[name: {}] not found for orgId: {} and serviceId: {}".format(group_name, org_id, service_id))
+        service_client = ServiceClient(self, service_metadata, group, service_stub, payment_channel_management_strategy(self), options)
```
chetankothari: I think we should accept an instance rather than instantiate it here. That way we don't have to worry about freezing the constructor api for the `PaymentChannelManagmentStrategy`.

vforvalerio87:
Hey, I was working on the CLI command to compile the client libraries for JS. I reviewed all the comments:
- I fixed the unused imports, am about to push a commit for that. Also fixed manifest, version.
- CLI setup.py now downloads and compiles the required node_module for generating JS client libraries. This is done in setup.py instead of the "blockchain install" script as the node_module (grpc-tools) has to be compiled for the CLI user's architecture so it cannot be built and packaged with the CLI distributable
- About the modules used in the CLI and SDK, I used two different virtual envs in my tests. I'll look to update the modules used in the SDK package too
- About the pricing and channel selection strategy: I straight up copied what's being done in the JS SDK for consistency so if we want to change how it works, we should do the same in JS. Agree we should have an abstracted "get_price_for_next_call".

raamb:
> * I fixed the unused imports, am about to push a commit for that. Also fixed manifest, version.

@vforvalerio87 have the commits been pushed?

vforvalerio87:
Set up monorepo:

this monorepo is organized as a PEP 420 compliant namespace
(https://www.python.org/dev/peps/pep-0420/).

The sdk package has been restructured to conform to the PEP 420
namespace package specification and it's under the "snet" namespace.

The cli package is both a regular package and a namespace package.

setup.py for the monorepo itself installs the dependencies for all
packages in the repository.

Each package and the overarching monorepo also have "requirements.txt"
files for local installations and CI enviroments: by installing via
"requirements.txt", the dependencies listed in "setup.py" are only
pulled and installed from PyPI if a local version is not already
available. This allows to get dependencies within the monorepo to be
installed from the local filesystem.

The whole package is called "snet" and gives the name to the namespace.
The subpackages are "snet.sdk", "snet.snet_cli" and snet_cli (since the
CLI package is both within the namespace and a regular package).
Going forward, all of the code in the "snet_cli" package can be migrated
to the namespace package: both were kept momentarily for compatibility.

In this scenario, everything inside the "snet.sdk" package has
visibility on the contents of the "snet.snet_cli" package and
vice-versa.
Everything inside the "snet_cli" regular package has visibility on the
contents of the "snet.snet_cli" package due to the regular Python import
mechanics, but it has no visibility on the contents of the "snet.sdk"
package (and vice-versa).
So, the shared code between the SDK and the CLI goes under the
"snet.snet_cli" package which is at "packages/snet_cli/snet/snet_cli".


Code and assets deduplication:

thanks to the new structure, everything under "resources" has been moved
to the "snet.snet_cli" package and is now shared between the SDK and
CLI. Specifically, this includes the json artifacts from the compilation
and deployment of the smart contracts, the raw ".proto" files to
interact with the daemon and the compiled python client libraries.


Changes:

since the CLI is both a namespace package and a regular package, changes
to the code have been kept to a minimum.
The only change of note is that state service does not compile its own
".proto" files at runtime inside the user's home directory anymore but
they are imported from the "resources" directory under the
"snet.snet_cli" namespace package, where they are compiled during setup.

vforvalerio87:
> Some of the suggestions were already pointed by others:
> 
> * Re-use blockchain/
> * Re-use scripts/

Done

> * test/ should go to the main project folder

I'd probably have each package have its own tests but we can discuss this

> * Maybe use just one .gitignore?

This is easier if stuff gets moved around since everything is relative to the package itself. More compartmentalized

> * setup.py:
>   
>   * python_requires='>=3.6',

Done

>   * url='https://github.com/singnet/snet-sdk-python', -> "monorepo"
> * MANIFEST.in:
>   
>   * snet_sdk/resources/proto/* doesn't exist.
>   * Re-use resources/ ?

This is dead now. All resources are in the cli package

> * VERSION (sdk gets it in a different way from cli).

Done

> * Unused imports (**init**.py):
>   - import operator
>   - from functools import reduce
>   - import sys
>   - import os
>   - import importlib
>   - import base64
>   - from pathlib import PurePath, Path
>   - import hashlib
>   - import collections
>   - from eth_account.messages import defunct_hash_message
>   - import snet_sdk.generic_client_interceptor as generic_client_interceptor

Fixed

> * **init**.py#L58
>   create_service_client(..., options={}) -> Mutable argument.
>   create_service_client(..., options=None)
>   if options is None:
>   options = dict()

Fixed

> * Suggestion (mpe_contract.py):
>   Maybe `EVENT_ABI` could be set like it is in `snet_cli/mpe_channel_command.py#L423`

Done

> * Unused arguments (service_client.py#L81):
>   `request_streaming` and `response_streaming`

`_intercept_call` is passed as an argument to `generic_client_interceptor.create`, which returns a `_GenericClientInterceptor` instantiated with `_intercept_call` as a parameter, its own `self._fn` equals `_intercept_call`, `self._fn` is always called with 4 parameters (+ self), so if you remove those two parameters from `_intercept_call`, the interceptor class is going to error out.

> * Remove compiled `state_service.proto` codes.

Done!

snet_sdk/snet_sdk/__init__.py
vforvalerio87: Done!

snet_sdk/scripts/blockchain
vforvalerio87: Done

vforvalerio87:
Obviously a lot more code can be shared and reused now, but I wanted to get the new repository structure understood and merged in first to serve as a basis for refactoring, especially to reuse the CLI code for dynamic compilation. Additional refactoring can come in later PRs

packages/sdk/snet/sdk/mpe_contract.py
```diff
@@ -0,0 +1,74 @@
...
+
+class MPEContract:
+    def __init__(self, w3):
+        self.web3 = w3
+        self.contract = get_contract_object(self.web3, "MultiPartyEscrow.json")
+        self.event_topics = [self.web3.sha3(text="ChannelOpen(uint256,uint256,address,address,address,bytes32,uint256,uint256)").hex()]
+        self.deployment_block = get_contract_deployment_block(self.web3, "MultiPartyEscrow.json")
+
+
+    def get_past_open_channels(self, account, service, starting_block_number=0, to_block_number=None):
```
astroseger: Why this function is so complicated? Why do you read logs in batches (while loop)??? Please take a look at ```MPEChannelCommand._get_all_filtered_channels``` in ```mpe_channel_command.py```. It works perfectly without while loop. You need to combine ```_get_all_channels_filter_sender_recipient_group``` and ```_get_all_filtered_channels```

packages/sdk/snet/sdk/service_client.py
```diff
@@ -0,0 +1,132 @@
...
+        if disable_blockchain_operations is False:
+            grpc_channel = self.grpc_channel
+        stub_instance = service_stub(grpc_channel)
+        return stub_instance
+
+
+    def _get_grpc_channel(self):
+        endpoint = self.options.get("endpoint", None)
+        if endpoint is None:
+            endpoint = next(filter(lambda endpoint: endpoint["group_name"] == self.group["group_name"], self.metadata["endpoints"]))["endpoint"]
```
astroseger: You can use get_endpoints_for_group  in MPEServiceMetadata

packages/sdk/snet/sdk/account.py
```diff
@@ -0,0 +1,88 @@
...
+            "chainId": int(self.web3.version.network),
+            "gas": DEFAULT_GAS, 
+            "gasPrice": self._get_gas_price(),
+            "nonce": self._get_nonce()
+        })
+        signed_txn = self.web3.eth.account.signTransaction(transaction, private_key=self.private_key)
+        return self.web3.toHex(self.web3.eth.sendRawTransaction(signed_txn.rawTransaction))
+
+
+    def send_transaction(self, contract_fn, *args):
```
astroseger: Why function for sending transaction is in Account class? In Account class, you mix two entities: 
- class for ETH/MPE account  (analog of MPEAccountCommand from snet-cli)
- class which sends on-chain transactions (analog of ContractCommand from snet-cli)
You probably should reuse contract.py from snet-cli and write better version of ContractCommand.

packages/sdk/snet/sdk/account.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+    def _get_gas_price(self):
+        return self.web3.eth.generateGasPrice()
+
+
+    def _send_signed_transaction(self, contract_fn, *args):
+        transaction = contract_fn(*args).buildTransaction({
+            "chainId": int(self.web3.version.network),
+            "gas": DEFAULT_GAS,
```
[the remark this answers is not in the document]
vforvalerio87: Removed, still works ok

raamb:
@arturgontijo @astroseger request your feedback on this. Thanks in advance

PR #255, opened 2019-06-04
Implement compilation of node.js client libraries

vforvalerio87:
Get node.js protoc compiler from npm on user's system (must be compiled
for the user's architecture, requires node.js to be used)
Allow users to compile client libraries for node.js

### DIFF
--- a/snet_cli/utils.py
+++ b/snet_cli/utils.py
@@ -148,15 +148,15 @@ def compile_proto(entry_path, codegen_dir, proto_file=None, target_language="pyt
             os.makedirs(codegen_dir)
         proto_include = pkg_resources.resource_filename('grpc_tools', '_proto')
 
-        protoc_args = [
-            "protoc",
+        compiler_args = [
             "-I{}".format(entry_path),
             "-I{}".format(proto_include)
         ]
 
         if target_language == "python":
-            protoc_args.append("--python_out={}".format(codegen_dir))
-            protoc_args.append("--grpc_python_out={}".format(codegen_dir)) 
+            compiler_args.insert(0, "protoc")
+            compiler_args.append("--python_out={}".format(codegen_dir))
+            compiler_args.append("--grpc_python_out={}".format(codegen_dir)) 
             compiler = protoc
         elif target_language == "nodejs":
             resources_path = Path(os.path.dirname(os.path.realpath(__file__))).joinpath("resources")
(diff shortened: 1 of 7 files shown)


snet_cli/identity.py
```diff
@@ -47,8 +47,8 @@ def __init__(self, w3, private_key):
                                                   curve=ecdsa.SECP256k1,
                                                   hashfunc=hashlib.sha256).get_verifying_key()
 
-        self.address = self.w3.toChecksumAddress(
-            "0x" + self.w3.sha3(hexstr=public_key.to_string().hex())[12:].hex())
+        self.address = self.w3.sha3(hexstr=public_key.to_string().hex())[12:].hex()
```
raamb: Is this related to the nodejs client stub generation?

snet_cli/identity.py
```diff
@@ -126,9 +126,9 @@ def __init__(self, w3, mnemonic, index):
         public_key = ecdsa.SigningKey.from_string(string=self.private_key,
                                                   curve=ecdsa.SECP256k1,
                                                   hashfunc=hashlib.sha256).get_verifying_key()
-
-        self.address = self.w3.toChecksumAddress(
-            "0x" + self.w3.sha3(hexstr=public_key.to_string().hex())[12:].hex())
+
```
raamb: Lets make this a method which can be called here as well as in the change above

snet_cli/identity.py
astroseger: Why have you removed toChecksumAddress?

vforvalerio87: Hey,

I did not remove toChecksumAddress: if you check the very next line it says:
```
self.address = self.w3.toChecksumAddress(self.address) if self.address.startswith("0x") else self.w3.toChecksumAddress("0x" + self.address)
```
I'm just checking if the generated `self.address` already starts with `0x` in which case I'm not adding it twice. Reason is that the `.hex()` method is already returning addresses prefixed with `0x` on my system so I get this error in the constructor of Identity:
```
ValueError: when sending a str, it must be a hex string. Got: '0x0x63fe11d4634044c2949623c48cc695b5cb0d14de'
```

which means I get this error all the time. This only started happening recently and I haven't looked into it yet much but at the same time, the change is not disruptive: if the outputted address already starts with `0x`, don't add it again.

vforvalerio87:
So, fun fact

I removed the change to `identity.py` to move it to its own separate PR, now CircleCI fails with my very same error:
```
++ snet account print
+ M0='Error: when sending a str, it must be a hex string. Got: '\''0x0xc0de46f2982b0a190ac93d659c5e0a41d6ff4352'\''
```

astroseger:
I found wired behavior of ```snet sdk generate-client-library```
```python
# will SILANTLY create "client_libraries" directory
snet sdk generate-client-library python snet example-service

# will complain that directory "output" is not exists
snet sdk generate-client-library python snet example-service output
```

It is small detail but I propose to fix it here (I will add it as small PR into @vforvalerio87 repository with tests)

snet_cli/utils.py
```diff
@@ -141,24 +142,41 @@ def get_cli_version():
...
 
-def compile_proto(entry_path, codegen_dir, proto_file=None):
+def compile_proto(entry_path, codegen_dir, proto_file=None, target_language="python"):
     try:
         if not os.path.exists(codegen_dir):
             os.makedirs(codegen_dir)
         proto_include = pkg_resources.resource_filename('grpc_tools', '_proto')
+
         protoc_args = [
             "protoc",
```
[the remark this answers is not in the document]
vforvalerio87: fair enough, done

PR #256, opened 2019-06-04
Ethereum address sanity check when generating from public key derived from private or mnemonic

vforvalerio87:
Verifying whether the address generated from the public key is already prefixed with 0x, in which case it's not added twice

### DIFF
--- a/snet_cli/identity.py
+++ b/snet_cli/identity.py
@@ -1,12 +1,11 @@
 import abc
 import json
-import hashlib
 import struct
 import time
 import getpass
 
+import web3
 from pycoin.key.BIP32Node import BIP32Node
-import ecdsa
 import rlp
 from eth_account.internal.transactions import serializable_unsigned_transaction_from_dict, encode_transaction, \
     UnsignedTransaction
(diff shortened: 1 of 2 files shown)


snet_cli/identity.py
```diff
@@ -47,8 +47,8 @@ def __init__(self, w3, private_key):
                                                   curve=ecdsa.SECP256k1,
                                                   hashfunc=hashlib.sha256).get_verifying_key()
 
-        self.address = self.w3.toChecksumAddress(
-            "0x" + self.w3.sha3(hexstr=public_key.to_string().hex())[12:].hex())
+        self.address = self.w3.sha3(hexstr=public_key.to_string().hex())[12:].hex()
+        self.address = self.w3.toChecksumAddress(self.address) if self.address.startswith("0x") else self.w3.toChecksumAddress("0x" + self.address)
```
astroseger: You probably should directly use web3.eth.accounts.privateKeyToAccount.

vforvalerio87: That's JavaScript: https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#privatekeytoaccount

I couldn't find the same in Python.

vforvalerio87:
> Am I right that you add this check because of some bug on your machine. Is it possible to add some tests (functional or unit) which cover this bug?

Actually this already fails for me with our current tests, see:
https://circleci.com/gh/singnet/snet-cli/278#tests/containers/0

vforvalerio87:
So apparently now this is a requirement for #255 

Tests already cover this in case it fails

Should we merge this in?

snet_cli/identity.py
astroseger: @vforvalerio87 it took 30sec to find web3py alternative!
```python
import web3
print(web3.eth.Account.privateKeyToAccount(self.private_key).address)
```
please use it

astroseger: (it is actually not so easy to find this function in web3py. I found it in the source code of web3py. But I think we should use it... )

PR #264, opened 2019-06-28
HOTFIX: some gas price strategies return floats but send_transaction demands an int. Casting to int

vforvalerio87:
Bumping version

Increasing timeout on waitForTransactionReceipt


packages/sdk/snet/sdk/account.py
```diff
@@ -59,7 +59,7 @@ def _send_signed_transaction(self, contract_fn, *args):
 
     def send_transaction(self, contract_fn, *args):
         txn_hash = self._send_signed_transaction(contract_fn, *args)
-        return self.web3.eth.waitForTransactionReceipt(txn_hash)
+        return self.web3.eth.waitForTransactionReceipt(txn_hash, 300)
```
astroseger: @raamb @vforvalerio87 
Are you sure that it is good solution? Why 300 is hardcoded?

astroseger: (In snet-cli we wait for transaction forever which might be not optiomal, but here we wait for 300 sec which will not be enough for mainnet)

vsbogd: Am I right that `send_transaction` is used internally to transfer tokens in automatic channel creation/updating strategy? If so I would make this timeout parameter of the strategy or at least configuration parameter.

vforvalerio87: Default for waitForTransactionReceipt in web3.py is 120 seconds, so I bumped it up in case the transaction is not mined within that timeframe.

I can just put the timeout as "None".

arturgontijo: I think we should keep the "waiting forever" but warn user that something is not right about the long waiting time. That way user will be able to know what's going on and CTRL+C the "Submitting transaction..." part.
Ref: [send_and_wait_for_transaction_receipt()](https://github.com/singnet/snet-cli/blob/master/packages/snet_cli/snet_cli/identity.py#L270)

vforvalerio87: I'm not sure why we are using the while + sleep instead of web3's `waitForTransactionReceipt` with no timeout to be honest

arturgontijo: I think 120s (no timeout) is not good enough for the Mainnet. With the while loop we can warn user every 60s (for example) that something is wrong with the network.

Another option (maybe better) is using the `waitForTransactionReceipt` inside the loop and when it raises the `Timeout Exception` (after 120s) we warn user and call `waitForTransactionReceipt` again.

PR #270, opened 2019-07-02
Dynamic compilation

vforvalerio87:
Added create_dynamic_serivce_client function

Dynamically generating a dynamic_service_client object by downloading and compiling protobuf for the user and generating ServiceClient instances for each stub


astroseger:
@vforvalerio87 
We still have 0 tests for SDK... I would prefer that we add some tests at least with this PR.

packages/sdk/snet/sdk/__init__.py
```diff
@@ -56,6 +59,47 @@ def create_service_client(self, org_id, service_id, service_stub, group_name=Non
         return service_client
 
 
+    def create_dynamic_service_client(self, org_id, service_id, group_name=None, payment_channel_management_strategy=PaymentChannelManagementStrategy, options=None):
```
astroseger: Could you give an example of using your SDK with dynamic classes? As usual better directly in the tests...

packages/snet_cli/snet/snet_cli/utils_proto.py
```diff
@@ -4,33 +4,42 @@
...
+            raise Exception("Error while loading protobuf. Found method %s.%s in multiple .proto files. We don't support packages yet!"%(service_name, method_name))
+        elif not service_name and method_name:
+            raise Exception("Error while loading protobuf. Found method %s in multiple .proto files. You could try to specify service_name."%method_name)
+        elif not service_name and not method_name:
+            return good_rez
+
     return good_rez[0]
 
-def _import_protobuf_from_file(grpc_pyfile, method_name, service_name = None):
+def _import_protobuf_from_file(grpc_pyfile, method_name=None, service_name=None):
```
astroseger: The output of ```import_protobuf_from_dir``` is completely different in case of ```method_name==None``` and ```method_name != None```. There are actually two functions with different behavior packed in one function which, in my opinion, complicate the code (the same for ```_import_protobuf_from_file```).

I suggest to write separate function (import_protobuf_from_dir_for_given_service_name ) and then reuse it in ```import_protobuf_from_dir```

packages/sdk/snet/sdk/__init__.py
vforvalerio87: Done, there's both tests and the new PR in the examples repo (https://github.com/singnet/snet-code-examples/pull/7)

vforvalerio87:
See latest commit:
added tests (also updated examples repo) + addressed comments

PR #368, opened 2019-11-27
Metadata changes in organization

pratik-vii:
Added organization type in organization metadata.
org_type argument has two choices "individual" and "organization".


vsbogd:
@pratik-vii , @keshrisohit , could you please explain it further? What is the difference between these two types of organizations from features perspective?

PR #373, opened 2019-12-03
added daemon addresses attribute in group level to the service metadata

pratik-vii:


vsbogd:
@pratik-vii, @prashantramangupta, could you please explain what is `daemon_addresses` value and how is it used? I cannot find any information and even usages in `singnet` repos.

PR #383, opened 2020-02-08
Add freecall support and introduce new payment strategy abstractions

keshrisohit:
Added Mpe abstraction
Added PaymentChannelProvider interface to interact with mpe contract
Added PaymenStartegy Abstraction
Added support for free call and add default FreeCallPaymentStrategy implementation


.circleci/config.yml
```diff
@@ -68,12 +68,12 @@ jobs:
 #        command: |
 #            sudo pip3 install nose
 #            nosetests -v --with-doctest
-    - run:
```
keshrisohit: will uncomment this

packages/snet_cli/snet/snet_cli/resources/proto/state_service.proto
```diff
@@ -2,6 +2,8 @@ syntax = "proto3";
```
anandrgitnirman: Would SUggest to download the proto from Daemon Artifact ( released version)

PR #392, opened 2020-03-12
increase transaction timeout, gas cost , and add token as hex

keshrisohit:


packages/sdk/README.md
```diff
@@ -38,6 +39,129 @@ from config import config
...
+
+To interact with SingularityNET services, you must compile the appropriate client libraries for that service.   
+To generate the client libraries to use in your application, you need the SingularityNET Command Line Interface, or CLI, which you can download from PyPi, see [https://github.com/singnet/snet-cli#installing-with-pip](https://github.com/singnet/snet-cli/snet_cli#installing-with-pip)
+  
+Once you have the CLI installed, run the following command:
+```bash
+snet sdk generate-client-library python <org_id> <service_id>
+```
+  
+Optionally, you can specify an output path; otherwise it's going to be `./client_libraries/python/<registry_address>/<org_id>/<service_id>`.
```
dagims: the location is at `./client_libraries/<org_id>/<service_id>/python`

packages/sdk/README.md
```diff
@@ -38,6 +39,129 @@ from config import config
...
+snet sdk generate-client-library python <org_id> <service_id>
+```
+  
+Optionally, you can specify an output path; otherwise it's going to be `./client_libraries/python/<registry_address>/<org_id>/<service_id>`.  
+You should move or copy these generated files to the root of your project.
+  
+Once you have installed the snet-sdk in your current environment and it's in your PYTHONPATH, you should import it and create an instance of the base sdk class:
+
+```python
+from snet_sdk import SnetSDK
```
dagims: this is the abandoned one right? the one from [singnet/snet-sdk-python](https://github.com/singnet/snet-sdk-python)?
in this module it is `from snet import sdk` ?

packages/sdk/README.md
```diff
@@ -38,6 +39,129 @@ from config import config
...
+
+---
+
+## Development
+
+### Installing
+
+#### Prerequisites  
+  
+* [Python 3.6.5](https://www.python.org/downloads/release/python-365/)
```
dagims: snet.sdk 0.2.1 requires python>=3.7 right?

packages/sdk/README.md
keshrisohit: Thanks for pointing this out. Have fixed all the comments.

PR #394, opened 2020-04-22
Increasing the gRPC message length to 1Gb.

arturgontijo:
To make `snet-cli` able to send and receive messages larger than `4Mb` (gRPC default).

PR #415, opened 2021-05-21
Increasing the gRPC message length to 1Gb

rajeevbbqq:
Conflict fix on #394


prashantramangupta (approved):
Approved. (As per @rajeevbbqq, @anandrgitnirman is aware of the change )

PR #418, opened 2021-05-24
Bug fix snet account deposit

rajeevbbqq:
Mint only if network is Local only

### DIFF
--- a/packages/snet_cli/snet_cli/commands/mpe_account.py
+++ b/packages/snet_cli/snet_cli/commands/mpe_account.py
@@ -31,7 +31,6 @@ class MPEAccountCommand(BlockchainCommand):
         already_approved = self.call_contract_command("SingularityNetToken", "allowance", [self.ident.address, mpe_address])
         if (already_approved < amount):
             self.transact_contract_command("SingularityNetToken", "approve", [mpe_address, amount])
-        self.transact_contract_command("SingularityNetToken", "mint", [self.ident.address, amount])
         self.transact_contract_command("MultiPartyEscrow", "deposit", [amount])
 
     def withdraw_from_mpe(self):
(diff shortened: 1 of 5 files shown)


packages/snet_cli/snet_cli/commands/mpe_account.py
```diff
@@ -41,5 +40,10 @@ def transfer_in_mpe(self):
         self.transact_contract_command("MultiPartyEscrow", "transfer", [self.args.receiver, self.args.amount])
 
     def mint_token(self):
-        address = self.args.address
-        self.transact_contract_command("SingularityNetToken", "mint", [address, int(self.args.amount)])
+        rez = self.config.session_to_dict()
```
anandrgitnirman, line 43: we nned to remove this definition here , not expose it as a command in snet-cli

packages/snet_cli/snet_cli/commands/mpe_account.py
```diff
@@ -41,5 +40,10 @@ def transfer_in_mpe(self):
...
+        network = rez['session']['network']
+
+        if network == 'local':
+            self.transact_contract_command("SingularityNetToken", "mint", [self.args.address, int(self.args.amount)])
```
anandrgitnirman, line 47: can we add this in the test case itself and remove referece to mint_token def in mpe_account.py

packages/snet_cli/snet_cli/commands/mpe_account.py
rajeevbbqq: Removed the cli command

packages/snet_cli/snet_cli/commands/mpe_account.py
rajeevbbqq: separated as test case

packages/snet_cli/test/functional_tests/mint/mint.py
```diff
@@ -0,0 +1,62 @@
+from packages.snet_cli.snet.snet_cli.utils.utils import get_web3, get_contract_object
+from web3.gas_strategies.time_based import medium_gas_price_strategy
+
+TRANSACTION_TIMEOUT = 500
+DEFAULT_GAS = 300000
+HTTP_PROVIDER = "http://localhost:8545"
+
+wallet_address_1 = "0x592E3C0f3B038A0D673F19a18a773F993d4b2610"
```
raamb, line 8: Lets remove private keys from the code.

anandrgitnirman: @rajeevbbqq , you can create a random pvt key and public key on the  fly , tools avb to do so

rajeevbbqq: @raamb & @anandrgitnirman hardcoded keys are also referred to in other test cases, and other test cases will fail if we create a key on the fly.

raamb: Ok sure. Lets go ahead with this for now

raamb (approved):

PR #435, opened 2022-06-23
SNET_SDK | Fixes channel processing issue

DhivakharVenkatachalam:
1. Get Ethereum log has a limit to filter logs up to 10000.
2. But we are having more than 10000 logs within the range of 20000 blocks.
3. Reducing the batch limit from 20000 to 5000.


prashantramangupta:
Pr is outdated, not required now.

PR #475, opened 2024-01-15
Change deprecated pkg_resources favor of importlib.metadata

kiruxaspb:
Changing deprecated `pkg_resources` favor of `importlib.metadata`

### DIFF
--- a/packages/sdk/setup.py
+++ b/packages/sdk/setup.py
@@ -1,4 +1,4 @@
-import pkg_resources
+import importlib.util
 from setuptools import setup, find_namespace_packages
 from os import path
 
@@ -10,10 +10,15 @@ PACKAGE_NAME = 'snet.sdk'
 
 
 def is_package_installed(package_name):
-    installed_modules = [p.project_name for p in pkg_resources.working_set]
-    print("Installed modules:")
-    print(installed_modules)
-    return package_name in installed_modules
+    spec = importlib.util.find_spec(package_name)
+    if spec is not None:
+        print(f"{package_name} is installed.")
+        return True
+    else:
+        print(f"{package_name} is not installed.")
+        return False
+
+
 
 
 dependencies = []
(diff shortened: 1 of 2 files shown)


packages/sdk/setup.py
```diff
@@ -10,10 +10,15 @@
 
 
 def is_package_installed(package_name):
-    installed_modules = [p.project_name for p in pkg_resources.working_set]
-    print("Installed modules:")
-    print(installed_modules)
-    return package_name in installed_modules
+    spec = importlib.util.find_spec(package_name)
```
sassless, line 13: According to the importlib [documentation](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec) this function
> Find the [spec](https://docs.python.org/3/glossary.html#term-module-spec) for a **module**, optionally relative to the specified package name.

So this is not a function that is suitable for use here.
Based on the message from [pkg_resources documentation](https://setuptools.pypa.io/en/latest/pkg_resources.html), it is advisable to use the [importlib_metadata](https://importlib-metadata.readthedocs.io/en/latest/) package here instead of importlib.
A more specific function from this package might be [importlib.metadata.metadata](https://importlib-metadata.readthedocs.io/en/latest/api.html#importlib_metadata.metadata).

sassless: Finally the whole function might be:
```python
import importlib.metadata  # importlib_metadata need to be installed


def is_package_installed(package_name):
    try:
        package = importlib.metadata.metadata(package_name)
        name, version = package.json["name"], package.json["version"]
        print(f"Installed {name} {version}")
        return True
    except importlib.metadata.PackageNotFoundError:
        print(f"Package {package_name} is not installed")
        return False
```

sassless (changes requested):

packages/sdk/setup.py
kiruxaspb: Resolved in [5db3daf](https://github.com/singnet/snet-cli/pull/475/commits/5db3dafb976e31a3f30d36452da1716938661edc)

kiruxaspb:
@sassless Updated

PR #490, opened 2024-04-08
Replaced the hardcoded contracts data with snet.contracts package import

PAY2109:


kiruxaspb (approved):

kiruxaspb (changes requested):
Need to add [snet.contracts](https://pypi.org/project/snet.contracts/) of the latest version (0.0.10) to the dependencies for building the package

PR #517, opened 2024-10-16
Predev channels caching

Arondondon:
In addition, "snet channel" section has changed, since the need for the concept of initialized channels has disappeared.


snet/cli/contract.py
```diff
@@ -22,6 +25,6 @@ def process_receipt(self, receipt):
 
         contract_events = map(lambda e: e["name"], filter(lambda e: e["type"] == "event", self.abi))
         for contract_event in contract_events:
-            events.extend(getattr(self.contract.events, contract_event)().process_receipt(receipt))
+            events.extend(getattr(self.contract.events, contract_event)().process_receipt(receipt, errors=DISCARD))
```
kiruxaspb, line 28: Here we ignoring UserWarnings from Web3 more about [here](https://web3py.readthedocs.io/en/stable/web3.contract.html#web3.contract.ContractEvents)

kiruxaspb (changes requested):
Need to add backward compatibility with configurations created by older versions of the CLI.

kiruxaspb:
> Need to add backward compatibility with configurations created by older versions of the CLI.

Added in https://github.com/singnet/snet-cli/pull/524

Issue #532, opened 2024-12-13
Implement PEP 625 – Filename of a Source Distribution


kiruxaspb:
Not so long ago, it was announced that the packages and file structure of libraries must now comply with the [PEP 625 standard](https://peps.python.org/pep-0625/).

We will need to consider this when releasing the next version, both here and in our SDK.

This will be needed to support the release of new versions for **PyPI**.

kiruxaspb:
Resolved in https://github.com/singnet/snet-cli/pull/534

PR #534, opened 2024-12-25
Fixed package name and reqs for implement pep 625

kiruxaspb:


kiruxaspb:
Resolved https://github.com/singnet/snet-cli/issues/532

